PR #63, opened 2024-09-16
Simplify Docker Image Build, be more similar to upstream build process

jj22ee:
*Issue #, if available:*

*Description of changes:*
- Build NPM tarball of ADOT JS in Docker Image and install tarball as a dependency of the `autoinstrumentation.ts` program, instead of copying all the source code alongside `autoinstrumentation.ts` as a non-dependency
- Simplify tsconfig.json usage, no need for base config.
- Upstream comparison:
  - https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/nodejs/package.json
  - https://github.com/open-telemetry/opentelemetry-operator/blob/main/autoinstrumentation/nodejs/Dockerfile

Testing:
- Tested docker image on ECS Fargate.


pxaws (approved):

Dockerfile
```diff
@@ -1,20 +1,20 @@
...
+COPY docker-utils/ .
 COPY aws-distro-opentelemetry-node-autoinstrumentation/tsconfig.json ./tsconfig.json
-COPY aws-distro-opentelemetry-node-autoinstrumentation/LICENSE ./LICENSE
-# ... but also add the required autoinstrumentation.ts and package.json to be consistent with upstream
-# https://github.com/open-telemetry/opentelemetry-operator/tree/main/autoinstrumentation/nodejs
-COPY docker-utils/autoinstrumentation.ts ./src/autoinstrumentation.ts
-COPY docker-utils/package.json ./package.json
-
+RUN cp /adot-js-build/aws-distro-opentelemetry-node-autoinstrumentation/aws-aws-distro-opentelemetry-node-autoinstrumentation-*.tgz ./
+RUN npm install aws-aws-distro-opentelemetry-node-autoinstrumentation-$(node -p -e "require('/adot-js-build/aws-distro-opentelemetry-node-autoinstrumentation/package.json').version").tgz
```
pxaws, line 17: Just curious: using "*" doesn't work here? That's why we need to specify the exact version?

jj22ee: I haven't tried just using `"*"` for npm install, but I want to parse the version as a sanity check to ensure we are installing the correct version from package.json.

PR #66, opened 2024-09-16
Update Lint Config to remove untyped Const warnings, run `npm run lint:fix`

jj22ee:
*Issue #, if available:*
- ~~Align with upstream, no need for typings for constants.~~
  - ~~See [upstream's `eslint.config.js`](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/b41797b2b8a11a7db4d3ec13c2452d824f45f82d/eslint.config.js#L43-L44)~~
- Not everyone adds typings for constants (which isn't required). This introduces many lint warnings.

*Description of changes:*
1. Update `eslint.config.js` typedef config
2. Run `npm run lint:fix` and generate new linted files

- Reduce lint warnings from 260 down to 6.
  - Before: [260 warnings](https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/10888963104/job/30214566328#step:5:328)
  - After: [6 warnings](https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/10913282764/job/30289505410#step:5:36)


aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/statistics.ts
```diff
@@ -8,7 +8,7 @@ export class Statistics implements ISamplingStatistics {
   public SampleCount: number;
   public BorrowCount: number;
 
-  constructor(requestCount: number = 0, sampleCount: number = 0, borrowCount: number = 0) {
```
pxaws, line 11: I feel like it make more sense to specify types here. We want people to know what parameters are expected without guessing. The same goes with function signatures. We want to rely on auto type inferring as much as possible but for those constructors, methods, functions, we want the type info as kind of documentation. 

Can we change the lint rule to enforce this? Maybe like this one: 

```
"@typescript-eslint/typedef": ["error", {
  "variableDeclaration": false,
  "parameter": true,
  "propertyDeclaration": true,
  "memberVariableDeclaration": true,
  "variableDeclarationIgnoreFunction": true
}],
```

Reference: https://typescript-eslint.io/rules/typedef/

jj22ee: I'll probably simplify the PR. I'll switch back to previous logic, but enforce errors on `typedef` lint config. I'll just need to update some missing types for constants.

pxaws (approved):

PR #118, opened 2024-11-06
feat: add support for new cohere command r models

yiyuan-he:
*Description of changes:*
Adding support for Cohere Command R models. The previous Cohere Command models are not yet fully deprecated ([EOL April 2025](https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html)) so we still include support for now.

Beginning 11/05/24 - Calls to old Cohere Command models now throw an exception for deprecation. I wasn't able to find any official announcement for this change, but I noticed it while testing during development in the Java SDK. 

Interestingly, calls to the old model still return a response so the full gen ai attributes are still generated for the time being.

Screenshot 2024-11-05 at 5 01 11 PM

*Test Plan:*
Verified the attributes for the Command R model is being generated with sample app auto-instrumentation.
Screenshot 2024-11-05 at 4 52 36 PM

### DIFF
--- a/aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
+++ b/aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
@@ -245,6 +245,22 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         if (requestBody.top_p !== undefined) {
           spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.top_p;
         }
+      } else if (modelId.includes('cohere.command-r')) {
+        if (requestBody.max_tokens !== undefined) {
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_MAX_TOKENS] = requestBody.max_tokens;
+        }
+        if (requestBody.temperature !== undefined) {
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TEMPERATURE] = requestBody.temperature;
+        }
+        if (requestBody.p !== undefined) {
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.p;
+        }
+        if (requestBody.message !== undefined) {
+          // NOTE: We approximate the token count since this value is not directly available in the body
+          // According to Bedrock docs they use (total_chars / 6) to approximate token count for pricing.
+          // https://docs.aws.amazon.com/bedrock/latest/userguide/model-customization-prepare.html
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_USAGE_INPUT_TOKENS] = Math.ceil(requestBody.message.length / 6);
+        }
       } else if (modelId.includes('cohere.command')) {
         if (requestBody.max_tokens !== undefined) {
           spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_MAX_TOKENS] = requestBody.max_tokens;
@@ -254,6 +270,9 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         }
         if (requestBody.p !== undefined) {
           spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.p;
+        }
+        if (requestBody.prompt !== undefined) {
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_USAGE_INPUT_TOKENS] = Math.ceil(requestBody.prompt.length / 6);
         }
       } else if (modelId.includes('ai21.jamba')) {
         if (requestBody.max_tokens !== undefined) {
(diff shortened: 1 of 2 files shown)


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -245,6 +245,22 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+        }
+        if (requestBody.p !== undefined) {
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.p;
+        }
+        if (requestBody.message !== undefined) {
+          // NOTE: We approximate the token count since this value is not directly available in the body
+          // According to Bedrock docs they use (total_chars / 6) to approximate token count for pricing.
+          // https://docs.aws.amazon.com/bedrock/latest/userguide/model-customization-prepare.html
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_USAGE_INPUT_TOKENS] = Math.ceil(requestBody.message.length / 6);
+        }
```
yiyuan-he, line 263: According to the [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-cohere-command-r-plus.html), this data should be available in the response body. However when logging the response body in the implementation it seems the data is not actually there. As a result, I decided to stay with this token approximation approach.


Screenshot 2024-11-06 at 9 38 29 AM

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -329,13 +348,18 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         if (responseBody.stop_reason !== undefined) {
           span.setAttribute(AwsSpanProcessingUtil.GEN_AI_RESPONSE_FINISH_REASONS, [responseBody.stop_reason]);
         }
-      } else if (currentModelId.includes('cohere.command')) {
-        if (responseBody.prompt !== undefined) {
+      } else if (currentModelId.includes('cohere.command-r')) {
+        console.log('Response Body:', responseBody);
+        if (responseBody.text !== undefined) {
           // NOTE: We approximate the token count since this value is not directly available in the body
           // According to Bedrock docs they use (total_chars / 6) to approximate token count for pricing.
```
yiyuan-he, line 355: The prompt is only available in the JavaScript implementation because of special data model defined in an upstream Otel package. This makes it possible to approximate the input token usage from the response body. 

However, this is not possible in the Java implementation as there is no special data model wrapping the inputs into the response body. As a result, I decided to move this approximation logic strictly to the request body to keep the implementation logic consistent between languages.

Screenshot 2024-11-06 at 9 42 25 AM

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -245,6 +245,22 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
...
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.p;
+        }
+        if (requestBody.message !== undefined) {
+          // NOTE: We approximate the token count since this value is not directly available in the body
+          // According to Bedrock docs they use (total_chars / 6) to approximate token count for pricing.
+          // https://docs.aws.amazon.com/bedrock/latest/userguide/model-customization-prepare.html
+          spanAttributes[AwsSpanProcessingUtil.GEN_AI_USAGE_INPUT_TOKENS] = Math.ceil(requestBody.message.length / 6);
+        }
       } else if (modelId.includes('cohere.command')) {
         if (requestBody.max_tokens !== undefined) {
```
yiyuan-he, line 265: I'm not sure if we should go ahead and remove support for the old `Cohere Command` model.

According to the [docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-lifecycle.html), EOL should not be until 2025 but we are already getting 404s to this model.

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -265,7 +284,7 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         if (requestBody.top_p !== undefined) {
           spanAttributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_TOP_P] = requestBody.top_p;
         }
-      } else if (modelId.includes('mistral.mistral')) {
+      } else if (modelId.includes('mistral')) {
```
yiyuan-he, line 287: We loosen this conditional because out of this [list](https://docs.aws.amazon.com/bedrock/latest/userguide/model-parameters-mistral-text-completion.html) for mistral models, one of the model ids starts with `mistral.mixtral` instead of `mistral.mistral`. The request/response body syntax is still the same for both.

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -329,13 +348,18 @@ export class BedrockRuntimeServiceExtension implements ServiceExtension {
         if (responseBody.stop_reason !== undefined) {
           span.setAttribute(AwsSpanProcessingUtil.GEN_AI_RESPONSE_FINISH_REASONS, [responseBody.stop_reason]);
         }
-      } else if (currentModelId.includes('cohere.command')) {
-        if (responseBody.prompt !== undefined) {
+      } else if (currentModelId.includes('cohere.command-r')) {
+        console.log('Response Body:', responseBody);
```
mxiamxia, line 352: pls remove all `info` level logging in SDK implementation. we don't want to flood customer's service logs

yiyuan-he: got it. removed this line

mxiamxia (approved):

PR #120, opened 2024-11-18
Support releasing layer with compatible runtimes node22.x

wangzlei:
*Issue #, if available:*

*Description of changes:*


.github/workflows/release-lambda.yml
mxiamxia, line 1: I suggest we also update the nodejs version in release workflow setup as well?
https://github.com/aws-observability/aws-otel-js-instrumentation/pull/120/files#diff-d28462c29540eddc84baa00bb9b0b1189782edfe56b84e600efc79a2d55cafb3R42

mxiamxia (approved):

PR #121, opened 2024-11-18
feat: [.NET and JS Feature Parity] Support for new AWS Resources in JS SDK

liustve:
*Description of changes:*

Adding support for new AWS resources in JS SDK. Part of an ongoing project to increase the ADOT SDK support in Node and .NET. 

Changes in this PR support the exact same features as the Python version: 
https://github.com/aws-observability/aws-otel-python-instrumentation/pull/265

Manual Testing:
Screenshot 2024-11-15 at 1 36 01 PM
Screenshot 2024-11-15 at 2 07 05 PM
Screenshot 2024-11-15 at 2 26 10 PM
Screenshot 2024-11-18 at 9 47 33 AM
Screenshot 2024-11-18 at 9 57 22 AM
Screenshot 2024-11-18 at 10 00 34 AM

Unit Tests for Instrumentation Patches and Metric Attribute Generators:

image


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/bedrock.ts
```diff
@@ -182,10 +183,15 @@ export class BedrockServiceExtension implements ServiceExtension {
...
+    const guardrailId = response.data[GUARDRAIL_ID];
+    const guardrailArn = response.data[GUARDRAIL_ARN];
 
-    if (guardrail_id) {
-      span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_BEDROCK_GUARDRAIL_ID, guardrail_id);
+    if (guardrailId) {
+      span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_BEDROCK_GUARDRAIL_ID, guardrailId);
+    }
+
+    if (guardrailArn) {
```
liustve, line 193: The guardrailArn attribute gets populated in the span for Python but the Javascript one does not. 

image

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/lambda.ts
```diff
@@ -0,0 +1,57 @@
...
+
+    let functionName = requestFunctionName;
+
+    if (requestFunctionName) {
+      if (requestFunctionName.startsWith('arn:aws:lambda:')) {
+        const split = requestFunctionName.split(':');
+        functionName = split[split.length - 1];
+      }
+
+      spanAttributes[AWS_ATTRIBUTE_KEYS.AWS_LAMBDA_FUNCTION_NAME] = functionName;
```
yiyuan-he, line 31: Previously, there was some bug with Lambda:Function:Name so we decided to temporarily remove support for it.

- https://github.com/aws-observability/aws-otel-python-instrumentation/pull/280
- https://github.com/aws-observability/aws-otel-python-instrumentation/pull/281

I'm not sure if that issue has been resolved yet. Should we remove it here as well?

cc: @mxiamxia

yiyuan-he:
I think this screenshot for Lamba Event Source Mapping is missing attributes:

image

liustve:
> I think this screenshot for Lamba Event Source Mapping is missing the new attributes:
> 
> image

Oh I need to edit how I'm populating the cloudformation attribute. I looked at the python code again and if there's no cfn then you just set it to the the resource identifier.

liustve:
> I think this screenshot for Lamba Event Source Mapping is missing the new attributes:
> 
> image

Fixed:

image

yiyuan-he (dismissed):

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/lambda.ts
jj22ee, line 1: Didn't take a look at this file, but we shouldn't have this file because upstream already has it.
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-aws-sdk-v0.44.0/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/lambda.ts

If we introduce our own lambda extension, we would be overwriting upstream's Lambda extension. We should be patching the existing extension instead, similarly like how we are doing this for SQS.
https://github.com/aws-observability/aws-otel-js-instrumentation/blob/v0.3.0/aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts#L108-L145

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/sns.ts
jj22ee, line 1: Same comment as the one I added for Lambda, we do not want to remove upstream's SNS extension:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-aws-sdk-v0.44.0/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services/sns.ts

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -56,6 +62,10 @@ describe('InstrumentationPatchTest', () => {
     expect(services.has('DynamoDB')).toBeTruthy();
     expect(services.has('Lambda')).toBeTruthy();
     // From patching but shouldn't be applied
```
jj22ee, line 64: Move SNS and Lambda from
- `// From patching but shouldn't be applied`

into
- `// Not from patching`

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -79,6 +89,10 @@ describe('InstrumentationPatchTest', () => {
     expect(services.has('DynamoDB')).toBeTruthy();
     expect(services.has('Lambda')).toBeTruthy();
     // From patching
```
jj22ee, line 91: Move SNS and Lambda into `// Not from patching`

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -57,6 +61,10 @@ export function applyInstrumentationPatches(instrumentations: Instrumentation[])
       if (services) {
         services.set('S3', new S3ServiceExtension());
         services.set('Kinesis', new KinesisServiceExtension());
+        services.set('Lambda', new LambdaServiceExtension());
+        services.set('SNS', new SNSServiceExtension());
+        services.set('SecretsManager', new SecretsManagerServiceExtension());
+        services.set('SFN', new StepFunctionsServiceExtension());
```
jj22ee, line 67: Please remove SNS and Lambda from this list.

Follow the SQS example instead:
`patchSqsServiceExtension(services.get('SQS'));`

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/aws/services/lambda.ts
liustve: Done. Ran a manual check to ensure the attributes still exists:

image
image

liustve: image

liustve: image

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
```diff
@@ -411,9 +450,18 @@ export class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
...
 
-    if (remoteResourceType !== undefined && remoteResourceIdentifier !== undefined) {
+    if (cloudFormationIdentifier === undefined) {
+      cloudFormationIdentifier = remoteResourceIdentifier;
+    }
+
+    if (
+      remoteResourceType !== undefined &&
+      remoteResourceIdentifier !== undefined &&
+      cloudFormationIdentifier !== undefined
```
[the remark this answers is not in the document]
liustve: Not sure, this is currently the same logic that's in python:

https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
```diff
@@ -411,9 +450,18 @@ export class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
       remoteResourceIdentifier = AwsMetricAttributeGenerator.getDbConnection(span);
     }
 
-    if (remoteResourceType !== undefined && remoteResourceIdentifier !== undefined) {
+    if (cloudFormationIdentifier === undefined) {
+      cloudFormationIdentifier = remoteResourceIdentifier;
```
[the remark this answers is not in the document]
liustve: Same as above^

aws-distro-opentelemetry-node-autoinstrumentation/test/aws-metric-attribute-generator.test.ts
```diff
@@ -806,10 +839,12 @@ describe('AwsMetricAttributeGeneratorTest', () => {
     validateRemoteResourceAttributes('AWS::Bedrock::DataSource', 'test_datasource_^^id');
     mockAttribute(AWS_ATTRIBUTE_KEYS.AWS_BEDROCK_DATA_SOURCE_ID, undefined);
 
-    // Validate behaviour of AWS_BEDROCK_GUARDRAIL_ID attribute, then remove it.
+    // Validate behaviour of AWS_BEDROCK_GUARDRAIL_ID and AWS_BEDROCK_GUARDRAIL_ARN attribute, then remove it.
```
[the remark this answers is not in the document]
liustve: Sorry, comment is a bit misleading. I believe since this is just asserting remote resource attributes and guardrail_arn is a cfa.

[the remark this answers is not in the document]
jj22ee: In that case, we should add test case for `AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER`. I noticed that there currently aren't any. I suppose we would add test cases for `AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER` into the `validateRemoteResourceAttributes()` test method, but I still want to get clarification if we always want to set this field.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
jj22ee: I think we need to get clarification if `cloudFormationIdentifier` always needs to be set, just in case it is also a bug in Python.

Otherwise, I think we should move the `cloudFormationIdentifier !== undefined` check nested inside this block instead (not outside of this block that I initially suggested).

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
jj22ee: Let's get clarification from team for this specification.

aws-distro-opentelemetry-node-autoinstrumentation/test/aws-metric-attribute-generator.test.ts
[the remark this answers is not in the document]
liustve: I can add the CFA tests into validateRemoteResourceAttributes() but would it be better to address the always setting CFA in a separate issue or PR? This PR just addresses the feature parity between JS and Python/Java.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
liustve: Sounds good!

aws-distro-opentelemetry-node-autoinstrumentation/test/aws-metric-attribute-generator.test.ts
```diff
@@ -1167,6 +1200,10 @@ describe('AwsMetricAttributeGeneratorTest', () => {
     testAwsSdkServiceNormalization('Kinesis', 'AWS::Kinesis');
     testAwsSdkServiceNormalization('S3', 'AWS::S3');
     testAwsSdkServiceNormalization('SQS', 'AWS::SQS');
+    testAwsSdkServiceNormalization('SNS', 'AWS::SNS');
+    testAwsSdkServiceNormalization('Lambda', 'AWS::Lambda');
+    testAwsSdkServiceNormalization('SecretsManager', 'AWS::SecretsManager');
+    testAwsSdkServiceNormalization('StepFunctions', 'AWS::StepFunctions');
```
jj22ee, line 1206: Can you verify that we should actually be testing that `SFN` (instead of `StepFunctions`) is normalized to `AWS::StepFunctions`? I see we commonly reference `SFN` in this PR. I suppose in that case, you need to update `normalizeRemoteServiceName` to handle `SFN` for this test to work.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-attribute-keys.ts
```diff
@@ -35,4 +36,12 @@ export const AWS_ATTRIBUTE_KEYS: { [key: string]: string } = {
   AWS_BEDROCK_KNOWLEDGE_BASE_ID: 'aws.bedrock.knowledge_base.id',
   AWS_BEDROCK_AGENT_ID: 'aws.bedrock.agent.id',
   AWS_BEDROCK_GUARDRAIL_ID: 'aws.bedrock.guardrail.id',
+  AWS_BEDROCK_GUARDRAIL_ARN: 'aws.bedrock.guardrail.arn',
+  AWS_SNS_TOPIC_ARN: 'aws.sns.topic.arn',
+  AWS_SECRETSMANAGER_SECRET_ARN: 'aws.secretsmanager.secret.arn',
+  AWS_STEPFUNCTIONS_STATEMACHINE_ARN: 'aws.stepfunctions.state_machine.arn',
+  AWS_STEPFUNCTIONS_ACTIVITY_ARN: 'aws.stepfunctions.activity.arn',
+  AWS_LAMBDA_FUNCTION_NAME: 'aws.lambda.function.name',
```
jj22ee, line 44: Seems this PR is currently missing the feature parity for setting `AWS_LAMBDA_FUNCTION_NAME`

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
jj22ee, line 1: I see some discrepancies against the Python implementation of `setRemoteResourceTypeAndIdentifier()`.

- Namely, `cloudFormationIdentifier` isn't set in cases `AWS_SQS_QUEUE_NAME`, `AWS_SQS_QUEUE_URL`, `AWS_BEDROCK_AGENT_ID`, and `AWS_BEDROCK_DATA_SOURCE_ID`.
- We are setting `cloudFormationIdentifier` in the `AWS_BEDROCK_GUARDRAIL_ID` case while Python does not.

Are these discrepancies expected?
Referencing: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/264

liustve: Yes, there will be another PR that sets AWS_SQS_QUEUE_NAME, AWS_SQS_QUEUE_URL, AWS_BEDROCK_AGENT_ID, and AWS_BEDROCK_DATA_SOURCE_ID CFAs.

It's not present in Michael's PR but the CFA is set in Python:

https://github.com/aws-observability/aws-otel-python-instrumentation/blob/main/aws-opentelemetry-distro/src/amazon/opentelemetry/distro/_aws_metric_attribute_generator.py

jj22ee (commented):
Thanks! I've left a few more minor comments.
Other than the previous unresolved comments, all else lgtm!

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
yiyuan-he: The CFN Id should always be set for every AWS resource otherwise the EMF Logs will not be correctly generated for AppSignals backend services.

IMO the original logic makes the most sense:
```
if (remoteResourceType !== undefined && 
    remoteResourceIdentifier !== undefined && 
    cloudFormationIdentifier !== undefined) {
    ...
}
```

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
yiyuan-he: [Related](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/121#discussion_r1866283099)

Agree, it probably doesn't make the most sense to set the CFN Id to connection string here. This condition was originally meant to default the CFN Id to the existing remote resource Id, as in most cases, they are the same. 

Unless this will cause some bug, I think the tradeoff here is one conditional for an extra unused attribute in DDB connection string vs. no conditional and manually setting the CFN for each case. 

Personally, I think the former is much more maintainable especially once we start supporting more AWS resources in the future. What do others think?

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
yiyuan-he: CFN Id for the Guardrail resource was addressed in different PR from the one referenced above: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/278

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
jj22ee: We can avoid manually setting the CFN for each case by catching all `AWS resource` cases with a `isAwsSDKSpan(span)` check (already has usage in this method). I think better be safe than sorry here.

See https://github.com/aws-observability/aws-otel-js-instrumentation/pull/121#discussion_r1866370956

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
yiyuan-he: Makes sense to me. We can go with this approach

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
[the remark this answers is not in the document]
yiyuan-he: Agree, I like this approach. Thanks for the suggestion/improvement!

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
```diff
@@ -406,6 +459,12 @@ export class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
           span.attributes[AwsSpanProcessingUtil.GEN_AI_REQUEST_MODEL]
         );
       }
+
+      if (cloudFormationIdentifier === undefined) {
+        cloudFormationIdentifier = remoteResourceIdentifier;
+      }
+
+      attributes[AWS_ATTRIBUTE_KEYS.AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER] = cloudFormationIdentifier;
```
jj22ee, line 467: Should this line be nested below in:
- `if (remoteResourceType !== undefined && remoteResourceIdentifier !== undefined) {`

I thought we originally wanted these 3 scenarios:
1. Set all 3: `AWS_REMOTE_RESOURCE_TYPE`, `AWS_REMOTE_RESOURCE_IDENTIFIER`, and `AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER`
2. Set 2:  `AWS_REMOTE_RESOURCE_TYPE`, `AWS_REMOTE_RESOURCE_IDENTIFIER`
3. Set `none`

With the current logic, a 4th scenario is possible:
- Set only: `AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER`

Do we want the 4th scenario? I imagine that `AWS_CLOUDFORMATION_PRIMARY_IDENTIFIER` must be dependent on the 2 other attributes (today we do not set either of those 2 attributes if one of those are missing)

mxiamxia (approved):

mxiamxia (approved):

PR #131, opened 2024-12-13
[Fix] Update AWS SDK Instrumentation to inject XRay trace context into HTTP Headers

jj22ee:
*Issue #, if available:*
Fixes the absence of broken X-Ray context propagation when the underlying HTTP instrumentation is suppressed or disabled.

Similarly to Java and Python, AWS SDK Js instrumentation itself should be able to inject the X-Ray Context into the HTTP Headers:
- [Java example](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/81c7713bb2f638c85006c3e152ad13d6e02f3259/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L308)
- [Python example
](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L163-L165)
- See Specification that clarifies that AWS SDK instrumentations should use X-Ray propagator specifically.
  - https://github.com/open-telemetry/opentelemetry-specification/blob/v1.40.0/supplementary-guidelines/compatibility/aws.md?plain=1#L9-L12

Note - If the underlying HTTP instrumentation is enabled, then the underlying HTTP Child Span of the AWS SDK Span will overwrite the Trace Context to propagate through headers.

*Description of changes:*
- Move patched/extended instrumentations to an `patches/extended-instrumentations/` directory
- Created `AwsSdkInstrumentationExtended` class that extends upstream AwsInstrumentation to override its patching mechanism of the `send` method. The overridden method will additionally update the AWS SDK middleware stack to inject the `X-Amzn-Trace-Id` HTTP header.


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/extended-instrumentations/aws-sdk-instrumentation-extended.ts
```diff
@@ -0,0 +1,42 @@
...
+// The overriden method will additionally update the AWS SDK middleware stack to inject the `X-Amzn-Trace-Id` HTTP header.
+//
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
+export class AwsSdkInstrumentationExtended extends AwsInstrumentation {
+  // Override the upstream private _getV3SmithyClientSendPatch method to add middleware to inject X-Ray Trace Context into HTTP Headers
+  // https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-aws-sdk-v0.48.0/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts#L373-L384
+  override _getV3SmithyClientSendPatch(original: (...args: unknown[]) => Promise<any>) {
+    return function send(this: any, command: V3PluginCommand, ...args: unknown[]): Promise<any> {
+      this.middlewareStack?.add(
```
mxiamxia, line 25: Q: is Middleware Stack existed on all AWS supported AWS JS SDK versions?  it learned it was introduced by AWS JS SDK V3 on 2020. 

https://aws.amazon.com/blogs/developer/modular-aws-sdk-for-javascript-release-candidate/

jj22ee: This PR will add support for v3.1.0
- `@aws-sdk/smithy-client ['^3.1.0']` (for aws-sdk packages before from `3.1.0` to `v3.363.0`)
- `@smithy/smithy-client ['>=1.0.3']` (for aws-sdk packages `v3.363.0+`)

>is Middleware Stack existed on all AWS supported AWS JS SDK versions
>learned it was introduced by AWS JS SDK V3 on 2020.

Yes, that is also near the beginning of AWS SDK JS V3 span. See version history:
- https://www.npmjs.com/package/@aws-sdk/client-s3?activeTab=versions
- https://www.npmjs.com/package/@aws-sdk/client-lambda?activeTab=versions
```
v3.1.0		54	4 years ago
v3.0.0		834	4 years ago
v1.0.0-rc.10	66	4 years ago
```

jj22ee: So there is no support for `v3.0.0`, but that's fine since [OTel doesn't support `v3.0.0` either](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/instrumentation-aws-sdk-v0.48.0/plugins/node/opentelemetry-instrumentation-aws-sdk/src/aws-sdk.ts#L85-L164)

mxiamxia (approved):
LGTM! Thanks

PR #132, opened 2024-12-13
[Fix] Update AWS SDK Instrumentation to inject XRay trace context into HTTP Headers

jj22ee:
- This is an alternative fix as opposed to https://github.com/aws-observability/aws-otel-js-instrumentation/pull/131

*Issue #, if available:*
Fixes the absence of broken X-Ray context propagation when the underlying HTTP instrumentation is suppressed or disabled.

Similarly to Java and Python, AWS SDK itself can inject the X-Ray Context into the HTTP Headers:
- [Java example](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/81c7713bb2f638c85006c3e152ad13d6e02f3259/instrumentation/aws-sdk/aws-sdk-2.2/library/src/main/java/io/opentelemetry/instrumentation/awssdk/v2_2/internal/TracingExecutionInterceptor.java#L308)
- [Python example
](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L163-L165)
- See Specification that clarifies that AWS SDK instrumentations should use X-Ray propagator specifically.
  - https://github.com/open-telemetry/opentelemetry-specification/blob/v1.40.0/supplementary-guidelines/compatibility/aws.md?plain=1#L9-L12

Note - If the underlying HTTP instrumentation is enabled, then the underlying HTTP Child Span of the AWS SDK Span will overwrite the Trace Context to propagate through headers.

*Description of changes:*
- Move patched/extended instrumentations to an `patches/extended-instrumentations/` directory
- Created `AwsSdkInstrumentationExtended` class that extends upstream AwsInstrumentation to add an additional instrumentation. The additional instrumentation will modify the `HttpRequest` class with an updated constructor to inject the `X-Amzn-Trace-Id` HTTP header.
  - Uses `propwrap.ts` that is taken from upstream OTel.


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/extended-instrumentations/aws-sdk-instrumentation-extended.ts
```diff
@@ -0,0 +1,42 @@
...
+  protected override init(): InstrumentationModuleDefinition[] {
+    const instrumentationModuleDefinitions = super.init();
+
+    const v3SmithyProtocolHttp = new InstrumentationNodeModuleDefinition(
+      '@smithy/protocol-http',
+      ['>=2.0.0'],
+      (moduleExports: any) => {
+        const newExports = propwrap(moduleExports, 'HttpRequest', (origHttpRequest: any) => {
+          class ExtendedHttpRequest extends origHttpRequest {
+            constructor(...args: any[]) {
```
mxiamxia, line 27: I like the other option than this one. The other is more cohesive to the `send` method itself and middleware is the right tool to modify the request from my view. :)

PR #149, opened 2025-02-05
Fix: Lambda Topology Issue

yiyuan-he:
**Issue #, if available:**
Lambda Topology Issue Context: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319

**Description of changes:**
- Apply fix for the Lambda Topology issue. The logic mimics the fix in our ADOT Python SDK.
- Adding back logic to generate `aws.lambda.function.name` and `aws.lambda.function.arn` span attributes for Lambda. Previously, we removed these changes in the AWS Resources expansion due to the Lambda Topology issue.
  - More context about adding support for Lambda as AWS Resource: 
    - https://github.com/aws-observability/aws-otel-python-instrumentation/pull/265

**Test plan:**

The same cases in the test plan of this [PR](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319) were validated by building a custom JavaScript lambda layer. Below is a screenshot of all the test cases in a single topology. 

We can observe the following:
- Service entity node for `Invoke` call to downstream lambda.
- AWS Resource node for `GetFunction` call to downstream lambda.
- AWS Resource node for `ListBuckets` call to downstream s3.
- Another AWS Resource node for `GetBucketLocation` call to downstream s3.

Screenshot 2025-02-05 at 10 15 26 AM

Here we see downstream lambda called with `Invoke` is correctly treated as RemoteService entity when not instrumented:
image

Additionally, I generated the spans locally to validate the new lambda instrumentation patch behaves correctly.

**Generated span for `Invoke` call**

We see the new `aws.remote.environment` attribute is present in span so the downstream lambda will be treated as a Service type in Application Signals backend.

invoke

**Generated span for `GetFunction` call**

We see `aws.remote.resource.identifier` and `aws.cloudformation.primary.identifier` are set so the downstream lambda will be treated as an AWS resource type in Application Signals backend.

get-function

### DIFF
--- a/aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
+++ b/aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
@@ -409,6 +409,33 @@ export class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
           this.extractResourceNameFromArn(activityArn)
         );
         cloudFormationIdentifier = AwsMetricAttributeGenerator.escapeDelimiters(activityArn);
+      } else if (AwsSpanProcessingUtil.isKeyPresent(span, AWS_ATTRIBUTE_KEYS.AWS_LAMBDA_FUNCTION_NAME)) {
+        // Handling downstream Lambda as a service vs. an AWS resource:
+        // - If the method call is "Invoke", we treat downstream Lambda as a service.
+        // - Otherwise, we treat it as an AWS resource.
+        //
+        // This addresses a Lambda topology issue in Application Signals.
+        // More context in PR: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319
+        //
+        // NOTE: The env vars LAMBDA_APPLICATION_SIGNALS_REMOTE_SERVICE and
+        // LAMBDA_APPLICATION_SIGNALS_REMOTE_ENVIRONMENT were introduced as part of this fix.
+        // They are optional and allow users to override the default values if needed.
+        if (AwsMetricAttributeGenerator.getRemoteOperation(span, SEMATTRS_RPC_METHOD) === 'Invoke') {
+          attributes[AWS_ATTRIBUTE_KEYS.AWS_REMOTE_SERVICE] =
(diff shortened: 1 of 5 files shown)


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -215,10 +221,41 @@ function patchLambdaServiceExtension(lambdaServiceExtension: any): void {
...
+        const requestFunctionNameFormat = request.commandInput?.FunctionName;
+        let functionName = requestFunctionNameFormat;
+
+        if (requestFunctionNameFormat) {
+          if (requestFunctionNameFormat.startsWith('arn:aws:lambda')) {
+            const split = requestFunctionNameFormat.split(':');
+            functionName = split[split.length - 1];
+          }
+          requestMetadata.spanAttributes[AWS_ATTRIBUTE_KEYS.AWS_LAMBDA_FUNCTION_NAME] = functionName;
+        }
```
yiyuan-he, line 234: Lambda API calls allow both shortened function name and full ARN to be passed as arguments. We conditionally extract the shortened function name here if the argument is in ARN format.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-metric-attribute-generator.ts
```diff
@@ -409,6 +409,33 @@ export class AwsMetricAttributeGenerator implements MetricAttributeGenerator {
...
+          }`;
+        } else {
+          remoteResourceType = NORMALIZED_LAMBDA_SERVICE_NAME + '::Function';
+          remoteResourceIdentifier = AwsMetricAttributeGenerator.escapeDelimiters(
+            span.attributes[AWS_ATTRIBUTE_KEYS.AWS_LAMBDA_FUNCTION_NAME]
+          );
+          cloudFormationIdentifier = AwsMetricAttributeGenerator.escapeDelimiters(
+            span.attributes[AWS_ATTRIBUTE_KEYS.AWS_LAMBDA_FUNCTION_ARN]
+          );
+        }
```
yiyuan-he, line 438: Mimic the logic to fix lambda topology issue from this PR: https://github.com/aws-observability/aws-otel-python-instrumentation/pull/319

PR #159, opened 2025-03-01
Remove old SigV4 headers before signing

liustve:
*Description of changes:*
There was a bug where the old headers were not cleaned before the request was signed causing 403 errors to occur when exporting traces to XRay OTLP endpoint.

### DIFF
--- a/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-aws-span-exporter.ts
+++ b/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-aws-span-exporter.ts
@@ -96,9 +96,10 @@ export class OTLPAwsSpanExporter extends OTLPProtoTraceExporter {
   // Need to ensure old SigV4 headers do not remain when we inject new SigV4 authorization headers.
   private removeSigV4Headers(headers: Record<string, string>) {
     const newHeaders: Record<string, string> = {};
+    const sigV4Headers = ['x-amz-date', 'authorization', 'x-amz-content-sha256', 'x-amz-security-token'];
 
     for (const key in headers) {
-      if (!key.toLowerCase().startsWith('x-amz-') && key.toLowerCase() !== 'authorization') {
+      if (!sigV4Headers.includes(key.toLowerCase())) {
         newHeaders[key] = headers[key];
       }
     }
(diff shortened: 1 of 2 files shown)


aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-aws-span-exporter.ts
```diff
@@ -93,6 +93,18 @@ export class OTLPAwsSpanExporter extends OTLPProtoTraceExporter {
     await super.export(items, resultCallback);
   }
 
+  // Need to ensure old SigV4 headers do not remain when we inject new SigV4 authorization headers.
+  private removeSigV4Headers(headers: Record<string, string>) {
+    const newHeaders: Record<string, string> = {};
+
+    for (const key in headers) {
+      if (!key.toLowerCase().startsWith('x-amz-') && key.toLowerCase() !== 'authorization') {
```
majanjua-amzn, line 101: I would say we should specifically write down the auth header keys instead of asserting `startsWith`. There are other headers that AWS provides in other use cases, like `x-amz-trace-id`, that can go into the headers in some cases.

liustve: Done!

majanjua-amzn (dismissed):

aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-aws-span-exporter.ts
```diff
@@ -93,6 +93,19 @@ export class OTLPAwsSpanExporter extends OTLPProtoTraceExporter {
     await super.export(items, resultCallback);
   }
 
+  // Need to ensure old SigV4 headers do not remain when we inject new SigV4 authorization headers.
+  private removeSigV4Headers(headers: Record<string, string>) {
+    const newHeaders: Record<string, string> = {};
+    const sigV4Headers = ['x-amz-date', 'authorization', 'x-amz-content-sha256', 'x-amz-security-token'];
+
+    for (const key in headers) {
+      if (!sigV4Headers.includes(key.toLowerCase())) {
```
pxaws, line 102: I am confused about the implementation and the comment: 
```
Need to ensure old SigV4 headers do not remain when we inject new SigV4 authorization headers.
```

Seems like we skip to set the headers for sigV4Headers?

liustve: The Sigv4 headers from previous exports are kept in the new request.
This causes the signer to use previously signed sigv4 headers when it injects the new headers here:
https://github.com/aws-observability/aws-otel-js-instrumentation/blob/87f1ffbe39b806d03e1a1d3a05bf8eab0e27776a/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-aws-span-exporter.ts#L82

pxaws (approved):

PR #161, opened 2025-03-04
Create NPM package for UDP Exporter

jj22ee:
*Issue #, if available:*
We want to make the OtlpUdpExporter class available as its own NPM package. Thus this PR is to [copy the implementation from ADOT](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-udp-exporter.ts) into its own UDP package.

*Description of changes:*
1. [copy the implementation from ADOT](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-udp-exporter.ts) into its own UDP package.
2. Ensure the release build workflow for ADOT Instrumentation only publishes ADOT Instrumentation (before, lerna will publish every package in the repo, which we don't want now since we now have >1 packages in this repo)

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -105,11 +105,12 @@ jobs:
              aws-distro-opentelemetry-node-autoinstrumentation/${{ env.ARTIFACT_NAME }} \
              ${{ env.ARTIFACT_NAME }}.sha256
 
-      # Publish to npm
-      - name: Publish to npm
+      # Publish '@aws/aws-distro-opentelemetry-node-autoinstrumentation' to npm
+      - name: Publish autoinstrumentation to npm
+        working-directory: aws-distro-opentelemetry-node-autoinstrumentation
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
           NPM_CONFIG_PROVENANCE: true
-        run: npx lerna publish from-package --no-push --no-private --no-git-tag-version --no-verify-access --yes
+        run: npm publish
     
     
--- a/exporters/aws-otel-otlp-udp-exporter/.eslintignore
+++ b/exporters/aws-otel-otlp-udp-exporter/.eslintignore
@@ -1,0 +1,5 @@
+build
+node_modules
+.eslintrc.js
+version.ts
+src/third-party
--- a/exporters/aws-otel-otlp-udp-exporter/.eslintrc.js
+++ b/exporters/aws-otel-otlp-udp-exporter/.eslintrc.js
@@ -1,0 +1,8 @@
+module.exports = {
+  "env": {
+    "commonjs": true,
+    "node": true,
+    "mocha": true,
+  },
+  ...require('../../eslint.config.js')
+}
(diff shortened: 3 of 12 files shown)


.github/workflows/release-build.yml
```diff
@@ -105,11 +105,12 @@ jobs:
...
-      # Publish to npm
-      - name: Publish to npm
+      # Publish '@aws/aws-distro-opentelemetry-node-autoinstrumentation' to npm
+      - name: Publish autoinstrumentation to npm
+        working-directory: aws-distro-opentelemetry-node-autoinstrumentation
         env:
           NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
           NPM_CONFIG_PROVENANCE: true
-        run: npx lerna publish from-package --no-push --no-private --no-git-tag-version --no-verify-access --yes
+        run: npm publish
```
pxaws, line 114: Any side effect by switching from `npx lerna` to `npm publish`? Does the option `--no-push --no-private --no-git-tag-version --no-verify-access --yes` apply to `npm publish`?

jj22ee: https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/README.md

> --[no-push](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/README.md)
By default, lerna version will push the committed and tagged changes to the configured [git remote](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/README.md#--git-remote-name). Pass --no-push to disable this behavior.

> --[no-private](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/README.md#--no-private)
By default, lerna version will include private packages when choosing versions, making commits, and tagging releases. Pass --no-private to disable this behavior.

> --[no-git-tag-version](https://github.com/lerna/lerna/blob/4fcac71c782cf91a3624fded775faaf3ddb8d636/libs/commands/version/README.md#--no-git-tag-version)
By default, lerna version will commit changes to package.json files and tag the release. Pass --no-git-tag-version to disable the behavior.

> --[no-verify-access](https://www.npmjs.com/package/@lerna/publish/v/5.4.0#--no-verify-access)
The legacy preemptive access verification is now off by default, so --no-verify-access is not needed. Requests will fail with appropriate errors when not authorized correctly. To opt-in to the legacy access verification, use [--verify-access](https://www.npmjs.com/package/@lerna/publish/v/5.4.0#--verify-access).

> --[yes](https://www.npmjs.com/package/@lerna/publish/v/5.4.0#--yes)
When run with this flag, lerna version will skip all confirmation prompts

None of these apply to [npm publish](https://docs.npmjs.com/cli/v8/commands/npm-publish), these are all lerna commands that are not needed.

pxaws (approved):

PR #173, opened 2025-04-02
Refactor Lambda Layer bundling

ezhang6811:
*Issue #, if available:*

*Description of changes:*
#171 introduced a bug where the Lambda server span was no longer produced in trace map for ESM-based handlers. This PR fixes the bug and now bundles the lambda layer directly instead of the autoinstrumentation package. The size of the bundled layer is now ~920KB, a nearly 94% reduction in size from the original 14.5MB.

The following files were copied from upstream to support Lambda layer bundling with minor changes: `webpack.config.js`, `tsconfig.webpack.json`, `tsconfig.esm.json`, and`install-externals.sh`.

Trace Map for CJS handler:
image

Trace Map for ESM handler:
image

Trace Map for ESM handler prior to fix (with bug):
image


lambda-layer/packages/layer/tsconfig.webpack.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "../../tsconfig.esm",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+  },
+  "include": [
+  "src/**/*.ts",
+  "test/**/*.ts"
```
jj22ee, line 9: nit - indents

lambda-layer/packages/layer/src/loader.mjs
jj22ee, line 1: I was wondering we needed the copyright headers [like this example](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/third-party/otel/aws/services/ServiceExtension.ts#L1-L15), but since the original files didn't have it, does it mean we don't need it?

jj22ee (commented):
Changes look good.

Can you list out which of these files are directly copied/modified from OTel's solution, and can you add images of the Trace map for both CJS and ESM Lambda handlers?

lambda-layer/packages/layer/tsconfig.webpack.json
ezhang6811: fixed

ezhang6811:
> Changes look good.
> 
> Can you list out which of these files are directly copied/modified from OTel's solution, and can you add images of the Trace map for both CJS and ESM Lambda handlers?

Description updated. I went back and removed the loader scripts since they weren't needed to instrument either handler.

jj22ee (dismissed):

lambda-layer/packages/layer/package.json
```diff
@@ -20,8 +20,8 @@
     "lint": "eslint . --ext .ts",
     "lint:fix": "eslint . --ext .ts --fix",
     "prepare": "npm run compile",
-    "compile": "tsc -p .",
-    "postcompile": "copyfiles -F 'node_modules/**' build/workspace/nodejs && copyfiles -f 'scripts/*' build/workspace && copyfiles -f 'build/src/*' build/workspace && cd build/workspace && bestzip ../layer.zip *"
+    "compile": "webpack && ./install-externals.sh",
```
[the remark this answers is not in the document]
ezhang6811: fixed

lambda-layer/packages/layer/install-externals.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -euf -o pipefail
+
+# Space separated list of external NPM packages
+EXTERNAL_PACKAGES=( "import-in-the-middle" )
```
mxiamxia, line 6: can you add sufficient comment to this file, why we need it and what is it for

ezhang6811: done

mxiamxia (approved):

PR #195, opened 2025-06-20
Auto-Configure ADOT SDK Defaults for Agent Observability

jj22ee:
*Issue #, if available:*

*Description of changes:*


aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
```diff
@@ -36,18 +37,68 @@ This file may also be used to apply patches to upstream instrumentation - usuall
...
-    process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS = 'fs,dns';
+    if (isAgentObservabilityEnabled()) {
+      // Assume users only need instrumentations that are manually set-up outside of OpenTelemetry
+      process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS =
+        'amqplib,aws-lambda,aws-sdk,bunyan,cassandra-driver,connect,cucumber,dataloader,dns,express,fastify,fs,generic-pool,graphql,grpc,hapi,http,ioredis,kafkajs,knex,koa,lru-memoizer,memcached,mongodb,mongoose,mysql2,mysql,nestjs-core,net,pg,pino,redis,redis-4,restify,router,socket.io,tedious,undici,winston';
+    } else {
+      // Disable the following instrumentations by default
+      // This auto-instrumentation for the `fs` module generates many low-value spans. `dns` is similar.
+      // https://github.com/open-telemetry/opentelemetry-js-contrib/issues/1344#issuecomment-1618993178
+      process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS = 'fs,dns';
```
mxiamxia, line 56: Q: why we don't check if there is any custom setting on this from customer first before setting the default? Also, I assume JS customers will use `OTEL_NODE_ENABLED_INSTRUMENTATIONS` to override this list?

jj22ee: 1. We do check for the custom setting (on line 47).
2. If an item exists on both `OTEL_NODE_DISABLED_INSTRUMENTATIONS` and `OTEL_NODE_ENABLED_INSTRUMENTATIONS`, it being disabled will take precedence. Also, if user sets the list inside `OTEL_NODE_ENABLED_INSTRUMENTATIONS`, then OTel JS will disable everything not included in the list.

PR #229, opened 2025-07-22
feat: Extract resource arn and remote resource access key for cross-account support

blairhyy-amazon:
*Description of changes:*
1. Add auto-instrumentation support for the following AWS resources.
- Populate `aws.stream.arn` in Span by extracting StreamArn from the
request body.
- Populate `aws.table.arn` in Span by extracting TableArn from the
response body.

2. Add auto-instrumentation support for remote resource access key.
- Populate `aws.auth.account.access_key` and `aws.auth.region` in Span
from STS credentials in client config
 
3. Generate cross-account metrics attributes when remote resource
identifier is present
- If remote resource arn is available, extract account id and region
from arn.
- Otherwise, pass account access key id and region from span to metric
if available.

4. Add unit tests, contract tests. Done E2E tests with CW agent. 


blairhyy-amazon:
Re-implement AWS SDK v3 Cross-Account Support with Recursion Fix

This PR reintroduces the AWS SDK v3 cross-account support (previously reverted in #192) with fixes for the [EKS E2E test failures](https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/16062476677/job/45752418669).

**Background**
The original implementation caused recursive instrumentation when running in EKS with IAM Roles for Service Accounts (IRSA). This occurred because IRSA uses the AWS STS AssumeRoleWithWebIdentity API ([ref](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html#irsa-oidc-background)), which triggers additional instrumented API calls during credential resolution.

**Issues Fixed**
1. Recursive instrumentation during credential resolution in IRSA environments
2. Repetitive middleware stack addition overrides to the same SDK client instance
3. Unnecessary span generation for internal credential-related calls

**Implementation Details**
1. Added context propagation with a credential capture skip flag for cross-account support
2. Implemented instrumentation suppression for internal credential extraction calls
3. Introduced `__adotMiddlewarePatched` guard to prevent duplicate middleware stack additions

**Testing**
- Verified E2E test pass in EKS environment with IRSA ([ref](https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/16434138225))
- Passed contract tests and unit tests 

**Related**
- Original PR: #192
- Revert PR: #212

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -364,67 +365,89 @@ const V3_CLIENT_CONFIG_KEY = Symbol('opentelemetry.instrumentation.aws-sdk.clien
...
-                  const region = await this.config.region();
-                  if (region) {
-                    span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_AUTH_REGION, region);
+          );
+
+          this.middlewareStack?.add(
+            (next: any, context: any) => async (middlewareArgs: any) => {
+              const activeContext = otelContext.active();
+              // Skip credential extraction if this is a nested call from another credential extraction
+              // This prevents infinite recursion when credential providers make AWS API calls
```
jj22ee, line 410: Can you add a unit test to validate the protection against infinite recursion?

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -364,67 +365,89 @@ const V3_CLIENT_CONFIG_KEY = Symbol('opentelemetry.instrumentation.aws-sdk.clien
 type V3PluginCommand = AwsV3Command<any, any, any, any, any> & {
   [V3_CLIENT_CONFIG_KEY]?: any;
 };
+// Symbol to prevent infinite recursion during credential capture
+// When we extract credentials, the AWS SDK may need to make additional AWS API calls
+// (e.g., sts:AssumeRoleWithWebIdentity) which go through the same instrumented 'send' method.
```
jj22ee, line 370: Just to clarify, the issue isn't because necessarily because the `send` method is instrumented. It is because the `middlewareStack` makes an AWS SDK call, which executes the `middlewareStack`, and so on. So what I mean is that for non-OTel and non-ADOT users, if they add the same `_adotExtractSignerCredentials` middleware into their own AWS SDK middleware stack, wouldn't they still see the same infinite recursion issue?

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -364,67 +365,89 @@ const V3_CLIENT_CONFIG_KEY = Symbol('opentelemetry.instrumentation.aws-sdk.clien
...
-                  if (region) {
-                    span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_AUTH_REGION, region);
+          );
+
+          this.middlewareStack?.add(
+            (next: any, context: any) => async (middlewareArgs: any) => {
+              const activeContext = otelContext.active();
+              // Skip credential extraction if this is a nested call from another credential extraction
+              // This prevents infinite recursion when credential providers make AWS API calls
+              if (activeContext.getValue(SKIP_CREDENTIAL_CAPTURE_KEY)) {
```
jj22ee, line 411: Just to be safe, I was wondering if there are more ways to safeguard against recursion, just in case we cannot fully rely on OTel Context for this?

Is it possible/safe to also set+unset some flag in the Client object in the beginning+end of your middleware, and then check the flag here?

```
if (activeContext.getValue(SKIP_CREDENTIAL_CAPTURE_KEY) || otherInfiniteRecursionSafeguardCheck) {
```

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -364,67 +365,89 @@ const V3_CLIENT_CONFIG_KEY = Symbol('opentelemetry.instrumentation.aws-sdk.clien
...
+                      if (credentials?.accessKeyId) {
+                        span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_AUTH_ACCOUNT_ACCESS_KEY, credentials.accessKeyId);
+                      }
+                    }
+                    if (this.config.region instanceof Function) {
+                      const region = await this.config.region();
+                      if (region) {
+                        span.setAttribute(AWS_ATTRIBUTE_KEYS.AWS_AUTH_REGION, region);
+                      }
+                    }
```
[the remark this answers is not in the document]
blairhyy-amazon: `accessKeyId` and `region` may vary due to role assumptions via AWS STS (Security Token Service). When assuming different IAM roles, the credentials (including accessKeyId) and target region can change dynamically.

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
blairhyy-amazon: When this middleware `_adotExtractSignerCredentials` executes, it calls `await client.config.credentials`. If credential resolution requires AWS API calls (such as STS AssumeRole operations), those API calls would themselves trigger the middleware again.

`middleware -> await client.config.credential -> triggers STS -> goes through middleware again -> ...`

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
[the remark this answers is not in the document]
jj22ee: To clarify, you mean for a single instantiated client object, the credentials/region can be changed in AWS SDK JS v3? I thought they are set per client, like:
```
      const lambda = new Lambda({
        region: region,
        credentials: {
          accessKeyId: 'abcde',
          secretAccessKey: 'abcde',
        },
      });
```
Can these be changed?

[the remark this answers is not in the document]
blairhyy-amazon: The credentials are set statically in your example. However, when using AWS STS, the user can provide temporary credentials via a credential provider that supports automatic rotation. 
```
const client = new LambdaClient({
  credentials: fromTemporaryCredentials({
    params: { RoleArn, RoleSessionName },
  }),
});
```
You are right that region and account id are fixed when the client is instantiated, but the access key ID and secret access key can be rotated automatically by the credential provider. Even though the access key ID changes with each credential refresh, it still belongs to the same assumed role.

aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
blairhyy-amazon: @jj22ee had a concern that if OTel, for some reason, doesn't use the expected Context when executing the AWS SDK call within the middleware, then in that case, the recursion mitigation would not work.

The current implementation should be safe because `otelContext.with(suppressedContext, async () => {...})` ensures that any code executed within its callback (including nested AWS SDK calls) will use the suppressed context where `SKIP_CREDENTIAL_CAPTURE_KEY` is set to true. As long as OTel's context propagation works as designed, this should prevent any recursion.

If not the case, we will remove the credential extraction middleware.

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -692,6 +692,61 @@ describe('InstrumentationPatchTest', () => {
...
+      // Mock OpenTelemetry context to track skip-credential-capture flag
+      const mockContext = {
+        getValue: (key: symbol) =>
+          key.toString().includes('skip-credential-capture') ? skipCredentialCaptureValue : undefined,
+        setValue: (key: symbol, value: any) => {
+          if (key.toString().includes('skip-credential-capture')) skipCredentialCaptureValue = value;
+          return mockContext;
+        },
+        deleteValue: () => mockContext,
+      };
```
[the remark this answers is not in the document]
blairhyy-amazon: Thank you. Updated.

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -692,6 +693,60 @@ describe('InstrumentationPatchTest', () => {
...
+      const middlewareStack: any[] = [];
+      const recursiveSdkSend = extractAwsSdkInstrumentation(PATCHED_INSTRUMENTATIONS)
+        ['_getV3SmithyClientSendPatch'](() => Promise.resolve())
+        .bind({
+          middlewareStack: { add: (middleware: any, config: any) => middlewareStack.push([middleware, config]) },
+          config: {
+            // Credentials provider that recursively calls SDK (simulates STS)
+            credentials: async () => {
+              credentialsProviderCallCount++;
+              await recursiveSdkSend({}, null);
```
jj22ee, line 725: Isn't this just the `Initial SDK call triggers middleware setup`? Shouldn't this call `await middlewareStack[1][0](() => Promise.resolve(), null)({});`?

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -692,6 +693,60 @@ describe('InstrumentationPatchTest', () => {
...
+      sinon.stub(trace, 'getSpan').returns(mockSpan as unknown as Span);
+
+      // Patch ROOT_CONTEXT to capture setValue calls
+      const originalSetValue = ROOT_CONTEXT.setValue;
+      ROOT_CONTEXT.setValue = (key: symbol, value: any) => {
+        if (key.toString().includes('skip-credential-capture')) {
+          skipCredentialCaptureValue = value;
+        }
+        return ROOT_CONTEXT;
+      };
```
jj22ee, line 712: This changes the actual functionality of `setValue` that we want to test - https://github.com/open-telemetry/opentelemetry-js/blob/48538b46add710483af3ff8147572d5bd4b273e0/api/src/context/context.ts#L50-L54.
However, the real functionality of `setValue` will return a new child Context, so your patch won't be applied to the child OTel Context. Let's just use the original `ROOT_CONTEXT` without patching it.

My understanding is that the `expect(credentialsProviderCallCount).toBe(1)` validation already implies that `expect(skipCredentialCaptureValue).toBe(true)` is validated. So maybe you don't need the `skipCredentialCaptureValue` validation?

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -692,6 +693,60 @@ describe('InstrumentationPatchTest', () => {
...
+            },
+            region: () => Promise.resolve('us-west-2'),
+          },
+        });
+
+      // Initial SDK call triggers middleware setup
+      await recursiveSdkSend({}, null);
+
+      // Execute credentials extraction middleware
+      await middlewareStack[1][0](() => Promise.resolve(), null)({});
```
jj22ee, line 736: As discussed, is it feasible to perform this `Execute credentials extraction middleware` through a realistic AWS SDK Client? I think you'd need to somehow patch the `config.credentials` method of the AWS SDK Client to perform another AWS SDK call.

This will also test that the mitigation applies when in a new Context, in which the middleware's AWS SDK call should be in.

aws-distro-opentelemetry-node-autoinstrumentation/test/patches/instrumentation-patch.test.ts
```diff
@@ -692,6 +693,41 @@ describe('InstrumentationPatchTest', () => {
...
+      const credentialsStsClient = new STS({ region: 'us-east-1' });
+
+      // Create main client with credentials provider that calls STS
+      const mainClient = new Lambda({
+        region: 'us-east-1',
+        credentials: async () => {
+          credentialsCallCount++;
+          // Simulate STS call for credentials (this should be skipped on recursion)
+          await credentialsStsClient.getCallerIdentity({}).catch((err: any) => {});
+          return { accessKeyId: 'sts-access-key', secretAccessKey: 'secret' };
```
jj22ee, line 709: `credentialsStsClient` doesn't have the custom `credentials`, only `mainClient` does. So there is no infinite recursion being done here.

Instead of Lambda, you can just use a single STS client that has your custom credentials. Then instead of `mainClient.invoke`, just do `credentialsStsClient.getCallerIdentity` so that there is possibility of infinite recursive `credentialsStsClient.getCallerIdentity` calls

blairhyy-amazon: Updated.

jj22ee (dismissed):
Approving, but I'm not 100% sure about the pure reliance on OTel's Context Management to prevent this [infinite recursion issue](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/229#issuecomment-3104843355). 

We've got a good unit test for this use case but I'm still a bit wary about not having a fallback mitigation for infinite recursion. Will rely on the @aws-observability/aws-application-signals-maintainers approver on whether or not we want to proceed with *only* this mitigation ([please see this comment thread](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/229#discussion_r2232368740)).

jj22ee (approved):
Re-approving since last review was stale

pxaws (approved):

PR #237, opened 2025-08-01
Add main build validation for release workflow

ezhang6811:
*Issue #, if available:*

*Description of changes:*
This PR modifies the release build workflow to wait for the main build workflow in the same branch to complete successfully. before proceeding with the release. see [Python PR](https://github.com/aws-observability/aws-otel-python-instrumentation/pull/443) for more details and testing.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -29,9 +29,11 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Check main build status
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           WORKFLOW_ID=$(gh api repos/${{ github.repository }}/actions/workflows --jq '.workflows[] | select(.name=="NodeJS Instrumentation Main Build") | .id')
-          LATEST_RUN=$(gh api repos/${{ github.repository }}/actions/workflows/$WORKFLOW_ID/runs --jq '.workflow_runs[] | select(.head_branch=="${{ github.ref_name }}") | {conclusion, status}' | head -1)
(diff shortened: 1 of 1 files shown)


.github/workflows/release-build.yml
```diff
@@ -21,7 +21,20 @@ permissions:
   contents: write
 
 jobs:
+  wait-for-main-build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Wait for E2E test to succeed
+        uses: ezhang6811/wait-on-check-action@6275ea93847ce5952ab8ceeac8fb1111820cf44d
```
jj22ee, line 28: Avoid using personal accounts for production release workflows. You could move the logic from the custom action to here if possible?

ezhang6811: fixed, replaced with local script

.github/workflows/release-build.yml
```diff
@@ -28,6 +28,24 @@ jobs:
       - name: Checkout Contrib Repo @ SHA - ${{ github.sha }}
         uses: actions/checkout@v4
 
+      - name: Check main build status
+        run: |
+          WORKFLOW_ID=$(gh api repos/${{ github.repository }}/actions/workflows --jq '.workflows[] | select(.name=="NodeJS Instrumentation Main Build") | .id')
+          LATEST_RUN=$(gh api repos/${{ github.repository }}/actions/workflows/$WORKFLOW_ID/runs --jq '.workflow_runs[] | select(.head_branch=="${{ github.ref_name }}") | {conclusion, status}' | head -1)
```
jj22ee, line 34: One assumption here is that this is ordered by date, which is likely true, but could(?) change in the future. Can you also sort the array from `workflow_runs[]` by the workflow date so guarantee the ordering so we can query the latest?

ezhang6811: good catch, updated in all 3 PRs. see latest commit for diff in jq command.

jj22ee (approved):

jj22ee (approved):

PR #253, opened 2025-08-18
Merge SDK and Lambda Releases

ezhang6811:
*Issue #, if available:*

*Description of changes:*
We plan to consolidate our ADOT SDK and Lambda layer releases for future versions. This PR merges the Lambda release workflow into the main release workflow, and publishes the layer artifacts and ARN notes to the same Github release as the SDK.

The release build workflow also now includes the SDK release notes with our upstream dependency versions to eliminate the manual effort needed when updating release notes.

Tested this workflow with all of the publishing jobs commented out: https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/17623772437/job/50075416236

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -28,7 +28,7 @@ permissions:
   contents: write
 
 jobs:
-  build:
+  build-sdk:
     environment: Release
     runs-on: ubuntu-latest
     steps:
@@ -74,7 +74,7 @@ jobs:
       # https://github.com/aws-observability/aws-otel-java-instrumentation/tree/93870a550ac30988fbdd5d3bf1e8f9f1b37916f5/smoke-tests
   
   build-layer:
-    needs: build
+    needs: build-sdk
     runs-on: ubuntu-latest
     outputs:
       aws_regions_json: ${{ steps.set-matrix.outputs.aws_regions_json }}
@@ -110,8 +110,8 @@ jobs:
           name: layer.zip
           path: lambda-layer/packages/layer/build/layer.zip
 
-  publish:
-    needs: [build, build-layer]
+  publish-sdk:
+    needs: [build-sdk, build-layer]
     runs-on: ubuntu-latest
     steps:
       - name: Checkout Repo @ SHA - ${{ github.sha }}
@@ -171,7 +171,7 @@ jobs:
 
   publish-layer-prod:
     runs-on: ubuntu-latest
-    needs: [build-layer, publish]
+    needs: [build-layer, publish-sdk]
     strategy:
       matrix:
         aws_region: ${{ fromJson(needs.build-layer.outputs.aws_regions_json) }}
@@ -255,7 +255,7 @@ jobs:
 
   generate-lambda-release-note:
     runs-on: ubuntu-latest
-    needs: [publish, publish-layer-prod]
+    needs: publish-layer-prod
     outputs:
       layer-note: ${{ steps.layer-note.outputs.layer-note }}
     steps:


wangzlei (dismissed):

wangzlei (dismissed):

.github/workflows/release-build.yml
```diff
@@ -116,20 +123,35 @@ jobs:
...
+          This release contains the following upstream components:
+
+          $DEPS
+
+          This release also publishes to public ECR and NPM.
+          * See ADOT node auto-instrumentation Docker image v${{ github.event.inputs.version }} in our public ECR repository:
+          https://gallery.ecr.aws/aws-observability/adot-autoinstrumentation-node
+          * See version ${{ github.event.inputs.version }} in our NPM repository:
+          https://www.npmjs.com/package/@aws/aws-distro-opentelemetry-node-autoinstrumentation
+          EOF
```
thpierce, line 146: In future, please do not combine features into one PR. It just makes it harder to review.

thpierce (commented):
A better solution would be:
1. Checkout/run Checks
2. Build everything inc. lambda
3. Create release notes (in memory only)
4. Publish everything, inc. lambda & release notes

This solution feels tacked-together and doesn't follow the basic tenet of: do the reversible things first (builds, checks, etc) then do irreversible things (publishing). 

What happens if there is a bug in creating notes or building Lambda? We are stuck in a bad state and now need to waste time trying to fix it.

.github/workflows/release-build.yml
```diff
@@ -105,37 +161,223 @@ jobs:
...
+          * See version ${{ github.event.inputs.version }} in our NPM repository:
+          https://www.npmjs.com/package/@aws/aws-distro-opentelemetry-node-autoinstrumentation
+
+          This release also includes the AWS OpenTelemetry Lambda Layer for JavaScript version ${{ github.event.inputs.version }}-$(echo $GITHUB_SHA | cut -c1-7).
+
+          Lambda Layer ARNs:
+          ${{ needs.generate-lambda-release-note.outputs.layer-note }}
+          EOF
+
+          shasum -a 256 ${{ env.ARTIFACT_NAME }} > ${{ env.ARTIFACT_NAME }}.sha256
```
thpierce, line 372: Why not `shasum -a 256 ${{ env.ARTIFACT_NAME }} | sed "s|aws-distro-opentelemetry-node-autoinstrumentation/||" > ${{ env.ARTIFACT_NAME }}.sha256`

.github/workflows/release-build.yml
```diff
@@ -105,37 +161,223 @@ jobs:
...
+          shasum -a 256 layer.zip > layer.zip.sha256
 
           gh release create --target "$GITHUB_REF_NAME" \
              --title "Release v${{ github.event.inputs.version }}" \
+             --notes-file release_notes.md \
              --draft \
              "v${{ github.event.inputs.version }}" \
-             aws-distro-opentelemetry-node-autoinstrumentation/${{ env.ARTIFACT_NAME }} \
+             ${{ env.ARTIFACT_NAME }} \
```
thpierce, line 380: So this works because we downloaded it in `Download SDK artifact`, above, right?

ezhang6811: yes, we don't need the full path anymore

.github/workflows/release-build.yml
ezhang6811: We downloaded the SDK artifact from a previous workflow as mentioned below. So we no longer have a prefix to remove from the file name when we create the SHA256 sum.

PR #274, opened 2025-09-30
Add NPM package for UDP Exporter

jj22ee:
*Issue #, if available:*
We want to make the UDP Exporter class available as its own NPM package. Thus this PR is to [copy the implementation from ADOT](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-udp-exporter.ts) into its own UDP package, with some renaming.

This PR was previously added but was removed in https://github.com/aws-observability/aws-otel-js-instrumentation/pull/174. So effectively, this PR reverts https://github.com/aws-observability/aws-otel-js-instrumentation/pull/174, but also includes some component renaming.

*Description of changes:*
1. [copy the implementation from ADOT](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/otlp-udp-exporter.ts) into its own UDP package.
2. Ensure the release build workflow for ADOT Instrumentation only publishes ADOT Instrumentation (before, lerna will publish every package in the repo, which we don't want now since we now have >1 packages in this repo)
3. Add test & release workflow for this component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,6 @@
   },
   "workspaces": [
     "aws-distro-opentelemetry-node-autoinstrumentation/",
-    "exporters/aws-distro-opentelemetry-exporter-xray-udp"
+    "exporters/aws-distro-opentelemetry-exporter-xray-udp/"
   ]
 }


wangzlei (approved):
LGTM

PR #289, opened 2025-10-25
Fix issue where UDP Exporter throws error in async callback, which isn't caught

jj22ee:
*Issue #, if available:*
In UDP Exporter, when sending via the socket, our callback method will rethrow the error if there an error from the `send`. However, this is done asynchronously, so it isn't actually caught in the surrounding `try/catch`.

This `err` can happen when the UDP Packet size limit is reached. When this error is thrown, it can cause the following error in Lambda:
```
[ERROR] [1761253646458] LAMBDA_RUNTIME Failed to post handler success response. Http response code: 403. 
{
    "errorMessage": "State transition from RuntimeResponseSentState to InvocationErrorResponse failed for runtime. Error: State transition is not allowed",
    "errorType": "InvalidStateTransition"
}
```

*Description of changes:*
- reject error, as opposed to throwing error because it cannot be caught in the existing surrounding `try/catch`.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -119,6 +119,16 @@ jobs:
       - name: Checkout Repo @ SHA - ${{ github.sha }}
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
+      - name: Build Tarball and Image Files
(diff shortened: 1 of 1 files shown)


Jasonlin1198:
Can someone help review this so that it can be included in the next layer release?

mxiamxia (approved):

PR #398, opened 2026-04-18
Lazy-load register.ts in tests to prevent TracerProvider conflicts

liustve:
### Description of changes
- Configurator, patch, and register tests statically imported `register.ts`, which calls `NodeSDK.start` at module load time, registering a global TracerProvider that conflicts with `contrib-test-utils` `registerInstrumentationTestingProvider`.
- Changed all three test files to lazy-load `register.ts` inside `before` hooks with `sinon.stub`.
- Added `propagation.disable` before `setGlobalPropagator` in patch tests so the XRay propagator registers correctly after `contrib-test-utils` sets a W3C-only propagator first.
- Fixes 2 pre-existing test failures in `customExtractor` and trace header injection tests.


jj22ee (approved):
lgtm!

wangzlei (approved):

PR #414, opened 2026-04-24
feat(agent-observability): register custom instrumentations in OTEL_NODE_ENABLED_INSTRUMENTATIONS default

liustve:
When `AGENT_OBSERVABILITY_ENABLED=true` and `OTEL_NODE_ENABLED_INSTRUMENTATIONS` is not set, the default value only included `aws-lambda,aws-sdk,http`. This caused our custom instrumentations (`aws_langchain`, `aws_openai_agents`, `aws_vercel_ai`) to be filtered out by `isInstrumentationDisabled()` since they were not in the enabled list.

This adds the three custom instrumentation short names to the default enabled list.


aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
```diff
@@ -75,7 +75,7 @@ export function setAwsDefaultEnvironmentVariables() {
       // instrumentations that are manually set-up outside of OpenTelemetry.
       // Our custom instrumentation short names are added after getNodeAutoInstrumentations()
```
syed-ahsan-ishtiaque: Does this needs to be updated as well?

syed-ahsan-ishtiaque (approved):

PR #418, opened 2026-04-25
feat: add undici to agent observability enabled instrumentations

liustve:
- Adds `undici` to the default `OTEL_NODE_ENABLED_INSTRUMENTATIONS` list when agent observability is enabled as`undici` is the default HTTP client in Node.js 18+ and is used by most agentic libraries for LLM API calls


liustve:
Replaced by clean PR off main

PR #432, opened 2026-05-08
pin @smithy/client to 4.12.13

ezhang6811:
*Issue #, if available:*

*Description of changes:*
https://github.com/smithy-lang/smithy-typescript/releases/tag/@smithy/smithy-client@4.13.0 breaks contract tests. Pin to previous version until upstream instrumentation changes entry hook.

### DIFF
--- a/contract-tests/images/applications/aws-sdk/package.json
+++ b/contract-tests/images/applications/aws-sdk/package.json
@@ -24,6 +24,7 @@
     "@aws-sdk/client-sns": "^3.750.0",
     "@aws-sdk/client-sqs": "^3.750.0",
     "@smithy/node-http-handler": "^4.0.0",
+    "@smithy/client": "4.12.13",
     "jszip": "^3.10.1",
     "node-fetch": "^2.7.0"
   }
--- a/contract-tests/images/applications/aws-sdk/package.json
+++ b/contract-tests/images/applications/aws-sdk/package.json
@@ -24,7 +24,7 @@
     "@aws-sdk/client-sns": "^3.750.0",
     "@aws-sdk/client-sqs": "^3.750.0",
     "@smithy/node-http-handler": "^4.0.0",
-    "@smithy/client": "4.12.13",
+    "@smithy/smithy-client": "4.12.13",
     "jszip": "^3.10.1",
     "node-fetch": "^2.7.0"
   }
--- a/contract-tests/images/applications/aws-sdk/package.json
+++ b/contract-tests/images/applications/aws-sdk/package.json
@@ -24,8 +24,10 @@
     "@aws-sdk/client-sns": "^3.750.0",
     "@aws-sdk/client-sqs": "^3.750.0",
     "@smithy/node-http-handler": "^4.0.0",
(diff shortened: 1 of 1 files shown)


contract-tests/images/applications/aws-sdk/package.json
```diff
@@ -24,6 +24,7 @@
     "@aws-sdk/client-sns": "^3.750.0",
     "@aws-sdk/client-sqs": "^3.750.0",
     "@smithy/node-http-handler": "^4.0.0",
+    "@smithy/smithy-client": "4.12.13",
```
srprash: If this is not a direct dependency of the application, then should this be pinned or overridden?
```
"overrides": {
    "@smithy/smithy-client": "4.12.13"
  }
```

srprash: #### Claude chat

Prompt: "https://github.com/aws-observability/aws-otel-js-instrumentation/pull/432/files
This PR addresses the contract test failure but is actually pins the dependency instead of overriding it. What is the difference between the two approaches and which one is better here and why?"

Response:
Here's the difference between the two approaches:

  PR #432's approach: Pin as direct dependency

  "@smithy/smithy-client": "4.12.13"
  Adds @smithy/smithy-client as a direct dependency with an exact version. npm will install this version at the top level of node_modules, and depending on npm's deduplication, other packages that need @smithy/smithy-client may use this
  version — but it's not guaranteed. If a transitive dependency (e.g., @aws-sdk/client-s3) requires ^4.12.13 and npm resolves the tree in a way that nests it, you could end up with two copies: 4.12.13 at the top and 4.13.0 nested inside
  the AWS SDK client. In that case, the OTel instrumentation might hook the top-level copy while the AWS SDK uses the nested one, or vice versa.

  Overrides approach

  "overrides": {
    "@smithy/smithy-client": "4.12.13"
  }
  This forces every transitive resolution of @smithy/smithy-client across the entire dependency tree to resolve to exactly 4.12.13. There will only ever be one copy, guaranteed.

  Which is better here?

  Overrides is more correct, because:

  1. Guarantees a single copy — no risk of the AWS SDK resolving a nested 4.13.0 while OTel hooks a different copy
  2. Explicit intent — it says "we are deliberately constraining a transitive dependency" rather than implying we directly use this package
  3. Doesn't pollute the dependency list with a package the application doesn't directly import

  That said, the PR #432 approach will likely work in practice for this contract test because the Docker container does a fresh npm install with a flat dependency tree (no lockfile committed), so npm will deduplicate and the pinned
  top-level version will satisfy all ^4.x ranges from the AWS SDK clients. It's a pragmatic fix even if not theoretically bulletproof.

  For a contract test image that's built fresh each time without a lockfile, either approach works. But overrides is the cleaner, more robust solution.

srprash: Doesn't matter much here but suggestion is to override.

ezhang6811: Good call out. Replaced with the override directive

srprash (approved):

PR #443, opened 2026-05-12
[feat] Add adaptive sampling with anomaly detection for AWS X-Ray Remote Sampler

Miqueasher:
## Background
Currently only Java and Python support Adaptive Sampling. This PR adds Adaptive Sampling to our JS SDK.

## Summary

- Add AdaptiveSamplingConfig parser supporting YAML/JSON config via `AWS_XRAY_ADAPTIVE_SAMPLING_CONFIG` env var (inline or file path)
- Add `AnomalyDetector` that identifies anomalous spans based on configurable conditions (error code regex, high latency threshold, specific operations)
- Add xrsr (X-Ray Sampling Rule) propagation via W3C tracestate — root spans set `xrsr=<hash>` of matched rule name, downstream services propagate upstream's hash unchanged. SHA-256 first 8 bytes, identical to Python/Java.
- Move boost stats to per-rule `SamplingRuleApplier` counters — prevents multi-counting across rules
- Split usage-type routing: `getAnomalyMatch()` returns {forBoost, forCapture} flags based on condition's usage field
- Use `AwsBatchUnsampledSpanProcessor` for anomaly capture (unsampled spans are dropped by regular BatchSpanProcessor)
- Root-span guard in `adaptSampling():` only count stats for spans whose sampling decision was made by this service
- Wire adaptive sampling into `AwsXRayRemoteSampler` with span batcher callback, threaded through `AwsSpanMetricsProcessor`
Address PR review comments: xrsr guard, usage routing, tests, and fixes

## Testing Done
- TypeScript compilation passes (npx tsc --noEmit — zero errors)
- 37 new unit tests added covering:
  - adaptive-sampling-config.ts — parsing, validation, usage types, edge cases
  - anomaly-detector.ts — condition matching, usage routing (forBoost/forCapture), capture logic
  - rule-cache-hash.ts — SHA-256 hash function, hash maps, getRuleApplierByHash, boost stats collection
  - sampling-rule-applier-boost.ts — countTrace dedup, countAnomalyTrace, snapshotBoostStatistics reset
- All 84 existing sampler tests pass — no regressions
- EC2 E2E with BoostVerificationTest sampling rule (SamplingRateBoost: {MaxRate: 0.5}):
  - traceState=xrsr=3023b7eacccb1e1a confirmed on all root spans
  - Per-rule SamplingBoostStatisticsDocuments reported consistently (TotalCount: 38-40, AnomalyCount: 19-20 per 10s window)
  - X-Ray responded with dynamic BoostRate increasing from 0.02 → 0.027
  - All stats accepted (UnprocessedBoostStatistics: [])
- Matches Python/Java behavior: same hash algorithm, same per-rule attribution, same X-Ray API contract


aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts
```diff
@@ -320,7 +341,7 @@ export class AwsOpentelemetryConfigurator {
     spanProcessors.push(new AwsBatchUnsampledSpanProcessor(spanExporter));
   }
 
-  static customizeSpanProcessors(spanProcessors: SpanProcessor[], resource: Resource): void {
+  static customizeSpanProcessors(spanProcessors: SpanProcessor[], resource: Resource, sampler?: Sampler): void {
     const baggageKeys: Set<string> = parseOtelBaggageKeysEnvVar();
```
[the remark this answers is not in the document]
hdj630: Is this related to adaptive sampling feature? Does java/python do the same?

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -138,19 +203,40 @@ export class _AwsXRayRemoteSampler implements Sampler {
   }
 
   private startSamplingTargetsPoller(): void {
-    // Update sampling targets every targetPollingInterval (usually 10 seconds)
     this.targetPoller = setInterval(
       () => this.getAndUpdateSamplingTargets(),
       this.targetPollingInterval * 1000 + this.targetPollingJitterMillis
     );
-    this.targetPoller.unref();
```
hdj630, line 146: This makes the timer hold the event loop open. What's the purpose of it?

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -104,20 +154,35 @@ export class _AwsXRayRemoteSampler implements Sampler {
...
-    if (matchedRule) {
-      return matchedRule.shouldSample(context, traceId, spanName, spanKind, attributes, links);
+    if (this.anomalyDetector) {
+      return {
+        decision: result.decision,
+        attributes: {
+          ...result.attributes,
+          [AWS_XRAY_ADAPTIVE_SAMPLING_CONFIGURED_ATTRIBUTE]: 'true',
+        },
+        traceState: result.traceState,
```
hdj630: The Java and Python ports both write the matched rule-name hash into W3C tracestate and baggage under key xrsr here, so downstream services can read it and attribute the trace's RequestCount /AnomalyCount / SampledAnomalyCount to the same rule that the upstream service matched.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -96,6 +118,34 @@ export class _AwsXRayRemoteSampler implements Sampler {
...
+  public setAdaptiveSamplingConfig(config: AdaptiveSamplingConfig): void {
+    this.anomalyDetector = new AnomalyDetector(config);
+    this.samplerDiag.info('Adaptive sampling enabled');
+  }
+
+  public setSpanBatcher(batcher: (span: ReadableSpan) => void): void {
+    this.spanBatcher = batcher;
+  }
+
+  public adaptSampling(span: ReadableSpan): void {
```
hdj630, line 141: Adaptive-sampling stats shoud not be counted on non-root spans — should be guarded.
adaptSampling runs from AwsSpanMetricsProcessor.onEnd for every span, including spans whose sampling decision was made upstream. Java (XrayRulesSampler.adaptSampling, lines 220–251 — comment "Span is not from an upstream service") and Python (_rule_cache.py::__get_effective_sampling_rule_applier, line ~198: not span.parent.is_valid) both skip countTrace / countAnomalyTrace when the span has a valid remote parent — unless the upstream xrsr resolves to a local rule. Otherwise this service inflates TotalCount / AnomalyCount / SampledAnomalyCount for traces it didn't actually sample, polluting X-Ray's boost stats.

1. Don't count adaptive-sampling stats for non-root spans (this method).
2. Don't write a fresh xrsr for non-root spans (Java XrayRulesSampler.shouldSample lines 152–159; Python _aws_sampling_result.py:39-40 — only set if parentTraceState.get(xrsr) == null).
3. Don't overwrite an existing xrsr from upstream.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/adaptive-sampling-config.ts
```diff
@@ -0,0 +1,155 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import { diag } from '@opentelemetry/api';
+import * as fs from 'fs';
+import * as yaml from 'js-yaml';
+
+export enum UsageType {
+  BOTH = 'both',
+  SAMPLING_BOOST = 'sampling-boost',
+  ANOMALY_TRACE_CAPTURE = 'anomaly_trace_capture',
```
hdj630: The AWS docs, Python (_aws_xray_adaptive_sampling_config.py), and Java (AwsXrayAdaptiveSamplingConfig.java) all use the hyphenated form:
 'anomaly-trace-capture' , not 'anomaly_trace_capture'

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
```diff
@@ -0,0 +1,150 @@
...
+  public snapshotAndResetStatistics(): BoostStatistics {
+    const snapshot = { ...this.stats };
+    this.stats = { TotalCount: 0, AnomalyCount: 0, SampledAnomalyCount: 0 };
+    return snapshot;
+  }
+
+  private matchesCondition(condition: AnomalyCondition, span: ReadableSpan): boolean {
+    if (condition.usage === UsageType.NEITHER) {
+      return false;
+    }
```
hdj630, line 90: We need to support sampling-boost only and anomaly-trace-capture only. With current implementation:
A user declaring usage: sampling-boost (intending "feed boost stats only, don't trigger anomaly capture") will still flow into the capture path via shouldCaptureAnomaly
A user declaring usage: anomaly-trace-capture (intending "capture only, don't inflate boost stats") will still bump AnomalyCount / SampledAnomalyCount via recordAnomaly.
Suggest changing isAnomaly and matchesCondition to return something like { matchesForBoost: boolean, matchesForCapture: boolean } (mirroring the Python/Java two-flag pattern), and routing each flag to the appropriate downstream call in _AwsXRayRemoteSampler.adaptSampling.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -138,19 +203,40 @@ export class _AwsXRayRemoteSampler implements Sampler {
...
+              AnomalyCount: boostStats.AnomalyCount,
+              SampledAnomalyCount: boostStats.SampledAnomalyCount,
+            });
+          }
+        }
+        if (boostDocs.length > 0) {
+          requestBody.SamplingBoostStatisticsDocuments = boostDocs;
+        }
+      }
+    }
```
hdj630, line 276: Same global BoostStatistics is copied into every rule's boost document — this multi-counts and mis-attributes traces. Stats should be counted per rule.

The current loop here pushes the same boostStats (global to the AnomalyDetector) into every non-Default rule that had RequestCount > 0. So if 3 rules are active, X-Ray sees every trace's anomaly counted 3× and attributed to rules it never matched, breaking the per-rule boost-rate computation.
 Two related issues at this site:
 
 1. Filtering on doc.RuleName !== 'Default' is JS-specific. Python suppresses by TotalCount > 0, not by rule name; Java reports Default like any other rule. Excluding Default also means a workload that only matches the catch-all default never reports any boost stats. Suggest dropping the !== 'Default' check.
 2. Per-rule attribution requires plumbing the matched rule down to the counting path. The natural fix is to move traces / anomalies / anomaliesSampled counters onto SamplingRuleApplier itself (mirroring Java/Python), have _AwsXRayRemoteSampler.adaptSampling resolve the effective applier (via local-root match or upstream xrsr), and call applier.countTrace() / applier.countAnomalyTrace() instead of the global AnomalyDetector counters. AnomalyDetector then becomes stateless w.r.t. boost stats and only owns anomaly detection + capture rate-limiting.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  public recordTrace(): void {
+    this.stats.TotalCount++;
+  }
+
+  public recordAnomaly(isSampled: boolean): void {
+    this.stats.AnomalyCount++;
+    if (isSampled) {
+      this.stats.SampledAnomalyCount++;
+    }
```
hdj630: BoostStatistics lives globally on AnomalyDetector — Python/Java keep these counters per-rule on the SamplingRuleApplier.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-span-metrics-processor.ts
```diff
@@ -95,6 +101,10 @@ export class AwsSpanMetricsProcessor implements SpanProcessor {
     for (const attribute in attributeMap) {
       this.recordMetrics(span, attributeMap[attribute]);
     }
+
+    if (this.sampler) {
+      this.sampler.adaptSampling(span);
```
hdj630, line 106: TotalCount incremented per span since you call this for every span end, but it should be per trace

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
```diff
@@ -0,0 +1,150 @@
...
+    // Rate limit
+    if (this.rateLimiter && !this.rateLimiter.take(1)) {
+      return false;
+    }
+
+    this.traceCache.set(traceId, true);
+    return true;
+  }
+
+  public recordTrace(): void {
```
hdj630: recordTrace needs to dedupe by trace ID before incrementing as totalCount should be per trace

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
```diff
@@ -0,0 +1,150 @@
...
+    const conditions = this.config.anomalyConditions;
+    if (!conditions || conditions.length === 0) {
+      return false;
+    }
+
+    // OR across conditions: any single condition match means anomaly
+    return conditions.some(condition => this.matchesCondition(condition, span));
+  }
+
+  public shouldCaptureAnomaly(traceId: string): boolean {
```
hdj630, line 72: Anomaly capture returns true only on the first span of an anomalous trace, then drops every follow-up span of the same trace. Actually we need capture all spans of an already-flagged trace, not only the anomaly span.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts
```diff
@@ -376,11 +397,38 @@ export class AwsOpentelemetryConfigurator {
...
+          resource
+        ).build();
+        const anomalyBatchProcessor = new BatchSpanProcessor(anomalyExporter);
+        spanProcessors.push(anomalyBatchProcessor);
+        xraySampler.setSpanBatcher(span => {
+          anomalyBatchProcessor.onEnd(span);
+        });
+        diag.info('Adaptive sampling anomaly capture enabled');
+      }
+
```
hdj630, line 419: Anomaly captures should go through AwsBatchUnsampledSpanProcessor, not a generic BatchSpanProcessor — and through the unsampled OTLP endpoint.

The default BatchSpanProcessor would normally drop those spans (it requires traceFlags & SAMPLED == 1).

Need to investigate why java/python can send out anomaly span with AwsBatchUnsampledSpanProcessor

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts
```diff
@@ -789,10 +850,12 @@ export class AwsSpanProcessorProvider {
       const configuredExporter: SpanExporter = AwsSpanProcessorProvider.customizeSpanExporter(exporter, this.resource);
       if (exporter instanceof ConsoleSpanExporter) {
         return new SimpleSpanProcessor(configuredExporter);
-      } else {
+      } else if (isLambdaEnvironment()) {
```
hdj630, line 847: Before this PR, all non-Console exporters used BatchSpanProcessor. This change downgrades every non-Lambda environment (EC2, ECS, EKS, on-prem, dev laptops) to SimpleSpanProcessor, which exports synchronously per span and meaningfully increases per-request latency and CPU under load. It applies regardless of whether AWS_XRAY_ADAPTIVE_SAMPLING_CONFIG is set, so every existing user of this distro who has adaptive sampling off gets the regression.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
hdj630: Is this change related to adaptive sampling project?

hdj630 (commented):
publish comments.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: Yes it was removed in the adaptive sampling project, when testing locally the `targetPoller` never ran while `.unref` was attached. I will find a solution.

Miqueasher: .unref() has been restored, it was a test env issue. It fires incorrectly for some reason when `OTEL_AWS_APPLICATION_SIGNALS_ENABLED` is not set to true.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts
Miqueasher: Restored `BatchSpanProcessor` for non-Lambda, non-Console environments.

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-opentelemetry-configurator.ts
Miqueasher: Replaced `BatchSpanProcessor` with `AwsBatchUnsampledSpanProcessor` for anomaly capture.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
Miqueasher: `shouldCaptureAnomaly` now returns true for repeat traceIds (trace already flagged for capture

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
Miqueasher: Same as above

aws-distro-opentelemetry-node-autoinstrumentation/src/aws-span-metrics-processor.ts
Miqueasher: `countTrace(traceId)` deduplicates by traceId via `seenTraceIds` Set (capped at 10K, cleared on snapshot).

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
Miqueasher: Counters now live on SamplingRuleApplier.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: Moved counters to per-rule SamplingRuleApplier (countTrace, countAnomalyTrace, snapshotBoostStatistics). AnomalyDetector is now stateless for boost stats. Removed !== 'Default' filter. Verified E2E: per-rule stats accepted by X-Ray with UnprocessedBoostStatistics: [].

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
Miqueasher: `isAnomaly()` replaced with `getAnomalyMatch()` returning {forBoost, forCapture}. adaptSampling routes each flag to the appropriate path: boost stats vs span capture.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/adaptive-sampling-config.ts
Miqueasher: Changed to 'anomaly-trace-capture'

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: Fixed. Matching Python `__get_effective_sampling_rule_applier` : xrsr lookup works for any span; local rule match only for spans with no valid parent. Child spans with local parent are skipped naturally (effectiveApplier resolves to undefined).

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: This was rebuilt to be identical to Pythons approach. `xrsr=<SHA256_hash_first_8_bytes>`

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -96,6 +118,48 @@ export class _AwsXRayRemoteSampler implements Sampler {
...
+
+    if (effectiveApplier) {
+      effectiveApplier.countTrace(traceId);
+    }
+
+    const match = this.anomalyDetector.getAnomalyMatch(span);
+    if (match) {
+      if (match.forBoost && effectiveApplier) {
+        effectiveApplier.countAnomalyTrace(isSampled);
+      }
```
hdj630: **Boost stats are pegged against every effective rule, not only boost-eligible ones — should add a `hasBoost()` gate.**

Java (`SamplingRuleApplier.java:148` + `XrayRulesSampler.java:281-290`) and Python (`_sampling_rule_applier.py:51` + `_rule_cache.py:186,192`) gate `countTrace` / `countAnomalyTrace` on `hasBoost()` / `has_boost()`, which is `true` only when the rule's **static** definition (from `GetSamplingRules`) contains a `SamplingRateBoost` block. The same gate covers both paths, because `effectiveApplier` is already resolved from either the upstream `xrsr` hash (downstream service) or the local match for root spans:

- **Root service:** count only if the matched rule has sampling boost.
- **Downstream service:** resolve the passed-in `xrsr` hash → local rule → count only if that rule has sampling boost.

This PR has no such gate, and `remote-sampler.types.ts` doesn't model the static `SamplingRateBoost` field on `ISamplingRule` at all — it only models the dynamic `SamplingBoost` returned by `GetSamplingTargets`. As written, boost stats are reported for any matched/resolved rule, polluting X-Ray's boost-decision input for rules that never opted into adaptive sampling.

Important: `hasBoost()` is **static eligibility** (`SamplingRateBoost != null`), not "a boost is currently active." Stats must be reported while the rule is merely eligible, so X-Ray has the data to decide whether to grant a boost (otherwise it's chicken-and-egg — no stats → no boost → no stats).

Suggested change — add the static field + accessor, then gate here:

```ts
// remote-sampler.types.ts — on ISamplingRule
SamplingRateBoost?: { MaxRate: number; CooldownWindowMinutes: number } | null;

// SamplingRuleApplier
public hasBoost(): boolean {
  return this.samplingRule.SamplingRateBoost != null;
}
```

```ts
// adaptSampling — single gate covers both root and downstream
if (effectiveApplier?.hasBoost()) {
  effectiveApplier.countTrace(traceId);
}

const match = this.anomalyDetector.getAnomalyMatch(span);
if (match) {
  if (match.forBoost && effectiveApplier?.hasBoost()) {
    effectiveApplier.countAnomalyTrace(isSampled);
  }
  if (match.forCapture && !isSampled && this.spanBatcher && this.anomalyDetector.shouldCaptureAnomaly(traceId)) {
    this.spanBatcher(span);
  }
}
```

(`xrsr` propagation in `shouldSample` can stay unconditional for any matched root rule — that matches Java `AwsSamplingResult.java:47-54` and Python `_aws_sampling_result.py:41-42`, which write `xrsr` gated only on the key not already being present, independent of boost eligibility.)

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
```diff
@@ -104,20 +168,58 @@ export class _AwsXRayRemoteSampler implements Sampler {
...
     }
 
-    const matchedRule: SamplingRuleApplier | undefined = this.ruleCache.getMatchedRule(attributes);
-    if (matchedRule) {
-      return matchedRule.shouldSample(context, traceId, spanName, spanKind, attributes, links);
+    // Determine xrsr tracestate value to propagate
+    // Python: _rule_cache.py lines 82-111
+    const parentSpanContext = trace.getSpan(context)?.spanContext();
+    const upstreamXrsr = parentSpanContext?.traceState?.get(XRSR_TRACE_STATE_KEY);
```
hdj630: **Missing baggage fallback when reading upstream `xrsr`.**

Java and Python read `xrsr` from **tracestate first, then fall back to baggage** if tracestate doesn't carry it:
- Java `XrayRulesSampler.java:152-162` — `parentSpanContext.getTraceState().get("xrsr")`, and if `null`, `Baggage.fromContext(parentContext).getEntryValue("xrsr")`.
- Python `_rule_cache.py:83-90` — `parent_span_context.trace_state.get("xrsr")`, and if `None`, `baggage.get_all(parent_context).get("xrsr")`.

This PR reads `xrsr` from `traceState` only (here at line 193, and also in `adaptSampling` at line 142) with no baggage fallback. A downstream JS service whose upstream carried `xrsr` in baggage (see below) will silently lose the rule attribution and fall back to local matching — diverging from Java/Python on the same trace.

Suggested: mirror the tracestate-first-then-baggage read in both `shouldSample` (this block) and `adaptSampling` (line 142):

```ts
import { propagation } from '@opentelemetry/api';

let upstreamXrsr = parentSpanContext?.traceState?.get(XRSR_TRACE_STATE_KEY);
if (!upstreamXrsr) {
  upstreamXrsr = propagation.getBaggage(context)?.getEntry(XRSR_TRACE_STATE_KEY)?.value;
}
```

**Why this fallback is real, not dormant — and where the producer side lives:**

The baggage read is the consumer side of a real bridge. The **AWS X-Ray propagator** (not the sampler) is what populates baggage: Java's `AwsXrayPropagator.inject()` reads `xrsr` from tracestate and writes it into the W3C `baggage` header, with the explicit comment *"Get sampling rule from trace state and inject into baggage / This is a back up in case the next service does not have trace state propagation."* So a Java upstream emits `xrsr` in both tracestate **and** baggage; a downstream that lost tracestate (proxies/LBs that strip it) still finds it in baggage.

Note the propagator-side bridge currently exists **only in Java** — the JS (`opentelemetry-js-contrib/propagator-aws-xray`) and Python X-Ray propagators don't do this copy yet. So:
- For **this** PR, adding the sampler-side baggage **read** (above) is the correct and sufficient scope — it matches Python and lets a JS service sitting downstream of a Java service still recover `xrsr`.
- The matching **producer** side for JS-origin traces (copy tracestate `xrsr` → baggage in `inject()`) is a separate change in `opentelemetry-js-contrib`'s `propagator-aws-xray`, out of scope here, but worth tracking so JS-origin traces get the same tracestate-stripping resilience Java already has.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/anomaly-detector.ts
```diff
@@ -0,0 +1,138 @@
...
+        }
+      }
+    }
+  }
+
+  public getAnomalyMatch(span: ReadableSpan): AnomalyMatch | null {
+    const conditions = this.config.anomalyConditions;
+    if (!conditions || conditions.length === 0) {
+      return null;
+    }
```
hdj630: **Missing the default anomaly-detection path — JS only honors explicit `anomalyConditions`, unlike Java and Python.**

When a boost-eligible rule has **no explicit `anomalyConditions`**, both Java and Python fall back to a *default* anomaly behavior: any span with HTTP status `> 499` (5xx) **or** span `StatusCode.ERROR` is treated as an anomaly (counted for boost + captured). This block returns `null` immediately when `anomalyConditions` is empty, so JS produces **zero** anomalies for such a rule — a behavioral divergence, not just a missing flag. A user who enables a boost on a rule without authoring explicit conditions gets adaptive sampling on Java/Python but a no-op on JS.

Reference implementations (they now agree):
- Python `_rule_cache.py:275-284` — `elif` (reached only when `anomaly_conditions` is falsy) triggers on `status_code > 499` or span `StatusCode.ERROR`.
- ADOT Java (via `.github/patches/opentelemetry-java-contrib.patch`, PR #1329 `feat: Allow disabling of default anomaly condition`, merged 2026-03-09, `b58e829`) — patch lines ~294-300: same `statusCode > 499` / `StatusCode.ERROR` default path.

**The opt-out flag goes with it.** That same default path is gated by `DisableDefaultAnomalyDetection`, a sub-field of the static `SamplingRateBoost` rule object:
- Python: parsed `_sampling_rule.py:17,22`, accessor `_sampling_rule_applier.py:144-148`, consumed `_rule_cache.py:279-282`.
- Java: added in #1329 — `getDisableDefaultAnomalyDetection()` + `isDefaultAnomalyDetectionDisabled()`, gating the default path (patch lines 38-50, 81-83, 123-124, 298-300).

So to match Java/Python, JS needs three coupled changes:
1. Model `SamplingRateBoost` on `ISamplingRule` (`remote-sampler.types.ts`), including `MaxRate`, `CooldownWindowMinutes`, and `DisableDefaultAnomalyDetection`. (This is the same modeling gap blocking the `hasBoost()` counting gate flagged in the other comment.)
2. Add a default-anomaly fallback here: when `anomalyConditions` is empty, treat `http.status_code`/`http.response.status_code > 499` or span `StatusCode.ERROR` as an anomaly (forBoost + forCapture).
3. Gate that fallback on the matched rule's `DisableDefaultAnomalyDetection` (skip the default path when the rule opts out).

Note `getAnomalyMatch` currently only takes the span — to honor the per-rule opt-out it'll need access to the effective rule's `DisableDefaultAnomalyDetection` (or the gate can live in `adaptSampling` where the effective applier is already resolved).

Miqueasher: `getAnomalyMatch` now accepts `defaultAnomalyDetectionDisabled` parameter and falls through to default 5xx/ERROR detection when no explicit conditions match. Gated on `isDefaultAnomalyDetectionDisabled()` per-rule. Added `DisableDefaultAnomalyDetection` to `SamplingRateBoost`

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: Baggage fallback has been implemented.

aws-distro-opentelemetry-node-autoinstrumentation/src/sampler/aws-xray-remote-sampler.ts
Miqueasher: Added `SamplingRateBoost` to `ISamplingRule`, `hasBoost()` on `SamplingRuleApplier`, and gated countTrace/countAnomalyTrace on `effectiveApplier?.hasBoost()`

PR #450, opened 2026-05-13
fix: add Converse API contract test and fix streaming responseHook return value

ezhang6811:
*Issue #, if available:*

*Description of changes:*
This PR adds a contract test case for the Bedrock Runtime Converse and ConverseStream APIs. This allows us to maintain support for gen_ai attributes and catch any changes in upstream support.

The patched responseHook was not returning the result from upstream's hook, which causes span.end() to not be called for streaming handlers. Added this fix to preserve responseHook value for streaming commands.


liustve (approved):

liustve:
non blocking but should add some more assertions in the future for other models

syed-ahsan-ishtiaque (approved):

PR #493, opened 2026-06-23
feat: feat: add OTel lite SDK for Lambda cold start optimization

wespil:
# otel-lite-sdk

A lightweight OpenTelemetry SDK replacement optimized for AWS Lambda cold starts. It replaces the full opentelemetry-sdk tracing pipeline with a minimal implementation that provides the same OTel API contract.

### Core tracing pipeline (opentelemetry-lite-sdk.ts):

- **TracerProvider** — implements TracerProviderAPI, holds a list of span processors and dispatches onStart/onEnd directly; `getTracer()` accepts `options.attributes` and forwards them to the instrumentation scope
- **Tracer** — implements TracerAPI, implements startSpan and startActiveSpan with parent-based always-on sampling
- **Span** — implements SpanAPI, supports setAttribute, setAttributes, addEvent, setStatus, recordException, updateName; guards all mutations against writes after end
- **InstrumentationScope** — minimal scope container with name, version, schemaUrl, and attributes (attributes are forwarded from `getTracer()` options and encoded into the OTLP output)
- **SpanEvent** — lightweight span event with name, attributes, and timestamp
- **SpanProcessor** — interface with onStart/onEnd/forceFlush/shutdown
- **BatchingSpanProcessor** — collects spans during invocation, computes local root status and propagates `faas.id` on start, exports all on forceFlush (no background threads or timers)
- **UdpSpanExporter** — encodes spans to OTLP protobuf and sends via UDP to the X-Ray daemon; injects Application Signals attributes when enabled
- **UdpExporter** — non-blocking UDP socket transport with X-Ray daemon protocol framing (T1S/T1U prefix + base64), socket is unref'd to not block Lambda shutdown
- **OTLP Protobuf Encoder** — hand-rolled encoder for ExportTraceServiceRequest (no protobuf library dependency); supports varint, fixed64, length-delimited wire types
  - `SpanStatus` uses correct protobuf field numbers: message → field 2, code → field 3
  - Span `flags` (field 16) sets `HAS_IS_REMOTE` (0x100) always, and `IS_REMOTE` (0x200) when the parent span context has `isRemote === true`
  - `InstrumentationScope` encodes scope attributes as field 3 key-value pairs
  - `scopeKey` uses null-byte separators and includes scope attributes to correctly group spans into distinct OTLP ScopeSpans entries
- **Resource** — built from OTEL_RESOURCE_ATTRIBUTES and OTEL_SERVICE_NAME env vars at init time (stored as Record<string, string>)
- **Parent-Based Always-On Sampling** — root spans are always sampled; child spans inherit traceFlags from the propagated parent context
- **Wall-clock-aligned timestamps** — calibrates `process.hrtime()` against `Date.now()` at module load for nanosecond-precision timestamps aligned with Lambda platform timing
- **Instrumentation Registry** — all 40 upstream `@opentelemetry/instrumentation-*` packages available via static thunks, selectable at runtime via `OTEL_NODE_ENABLED_INSTRUMENTATIONS` (default: `aws-sdk,http,aws-lambda`)

## Activating Fast Start Mode

Add a new environment variable: `OTEL_AWS_LAMBDA_FAST_START=true`

## What's supported vs. the standard SDK path

| Feature / Env Var | Standard SDK | Fast Start (Lite) | Behavior in Fast Start |
|---|---|---|---|
| `OTEL_RESOURCE_ATTRIBUTES` | Honored | Honored | Parsed at init time |
| `OTEL_SERVICE_NAME` | Honored | Honored | Overrides `service.name` from resource attrs |
| `AWS_XRAY_DAEMON_ADDRESS` | Not used | Honored | UDP endpoint for trace export |
| `OTEL_AWS_APPLICATION_SIGNALS_ENABLED` | Honored | Honored | Injects aws.local/remote attributes |
| `OTEL_NODE_ENABLED_INSTRUMENTATIONS` | Honored | Honored | Selects which instrumentations to activate |
| `OTEL_NODE_DISABLED_INSTRUMENTATIONS` | Honored | Honored | Subtracts from the enabled list |
| `OTEL_LOG_LEVEL` | Honored | Honored | Wired to diag logger for debug output |
| `OTEL_PROPAGATORS` | Honored | Silently ignored | Hardcoded: xray + tracecontext + baggage |
| `OTEL_TRACES_SAMPLER` | Honored | Silently ignored | Always samples root spans |
| `OTEL_TRACES_EXPORTER` | Honored | Silently ignored | Always UDP to X-Ray daemon |
| `OTEL_EXPORTER_OTLP_ENDPOINT` | Honored | Silently ignored | Uses `AWS_XRAY_DAEMON_ADDRESS` |
| `OTEL_METRICS_EXPORTER` | Honored | No metrics support | Silently ignored |
| `OTEL_LOGS_EXPORTER` | Honored | No logs support | Silently ignored |
| Custom SpanProcessor | Supported | Not supported | Only built-in BatchingSpanProcessor |
| Custom Exporter | Supported | Not supported | Only UDP to X-Ray daemon |
| Span Links | Encoded | Accepted but dropped | Parameter kept for API compatibility |

### Bootstrap (register.ts):

- Detects `OTEL_AWS_LAMBDA_FAST_START=true`, calls `configureLiteMode()` which:
  - Instantiates TracerProvider + BatchingSpanProcessor + UdpSpanExporter
  - Registers as the global TracerProvider via `trace.setGlobalTracerProvider()`
  - Sets up AsyncLocalStorageContextManager as the global context manager
  - Sets up composite propagator (Baggage + X-Ray + W3C TraceContext)
  - Builds and enables instrumentations via the registry (`buildInstrumentations()`)
  - Applies the shared smithy-send-patch to the aws-sdk instrumentation
- Then returns early, skipping the full ADOT SDK initialization

### Lambda wrapper (otel-instrument):

- Defaults `OTEL_AWS_LAMBDA_FAST_START` to `"false"` if not set
- When `OTEL_AWS_LAMBDA_FAST_START=true`: uses `--require /opt/lite-wrapper.js`, bypassing `register.ts` and `import-in-the-middle` entirely
- When `OTEL_AWS_LAMBDA_FAST_START=false`: falls back to the standard `--require /opt/wrapper.js` (CJS) or `--import /opt/wrapper.mjs` (ESM) path
- Defaults `OTEL_NODE_ENABLED_INSTRUMENTATIONS` to `aws-sdk,aws-lambda,http` if unset

### Application Signals support:

- When `OTEL_AWS_APPLICATION_SIGNALS_ENABLED=true`, the exporter injects `aws.local.service`, `aws.local.operation`, `aws.local.environment`, `aws.remote.service`, `aws.remote.operation`, `aws.span.kind`, and `aws.is.local.root` attributes onto spans before OTLP encoding
- When disabled, `aws.is.local.root` is stripped from exported attributes to match the full SDK's output format
- Spans are sent via UDP with T1S (sampled) or T1U (unsampled) prefix based on the parent's sampling decision
- Service operations and dependencies appear in Application Signals without requiring EMF metrics or the ADOT collector

### AWS SDK instrumentation (patches/smithy-send-patch.ts):

- Shared between the lite SDK and the full SDK (`instrumentation-patch.ts`), gated by `OTEL_AWS_LAMBDA_FAST_START`
- Overrides `_getV3SmithyClientSendPatch` to add three middlewares on first `send()` call:
  - **_adotInjectXrayContextMiddleware** (build step) — injects X-Ray trace context into outgoing request headers
  - **_adotExtractCredentials** (build step, fast start) / **_adotExtractSignerCredentials** (build step, standard) — captures `aws.auth.account.access_key` and `aws.auth.region` from the client config. Standard mode uses `suppressTracing` + recursion guard; fast start uses simple try/catch
  - **_adotCaptureResponseMetadata** (deserialize step) — captures `aws.request.id`, `aws.request.extended_id`, and `http.status_code` from `$metadata` on the response
- Uses `factoryArgs.find(typeof === 'function')` to robustly resolve the original `send` regardless of the upstream's bound-argument signature

### Attribute propagation (BatchingSpanProcessor.onStart):

- Computes `aws.is.local.root` by checking if the parent span context is absent, invalid, or remote
- Propagates `faas.id` from parent to child spans when not already present

### Instrumentation Registry (buildInstrumentations):

- Maps all 40 upstream instrumentation packages as static `() => require('literal').Class` thunks
- Webpack bundles all 40 into `lite-wrapper.js` (sharing a single `@opentelemetry/api` instance), ensuring spans flow on Lambda where the function ships its own api copy
- Selection at runtime via `OTEL_NODE_ENABLED_INSTRUMENTATIONS` / `OTEL_NODE_DISABLED_INSTRUMENTATIONS` — same env-var contract as the full SDK's `getNodeAutoInstrumentations()`
- Default exclusions: `fs` (too noisy) and `host-metrics` (metrics-only, no trace output)
- Only enabled instrumentations are instantiated; disabled ones incur zero runtime cost


aws-distro-opentelemetry-node-autoinstrumentation/src/patches/instrumentation-patch.ts
```diff
@@ -456,105 +452,9 @@ function patchAwsLambdaInstrumentation(instrumentation: Instrumentation): void {
...
-            {
-              step: 'build',
-              name: '_adotExtractSignerCredentials',
-              override: true,
-            }
-          );
-          this.__adotMiddlewarePatched = true;
-        }
+import { applySmithySendPatch, SKIP_CREDENTIAL_CAPTURE_KEY } from './smithy-send-patch';
+export { SKIP_CREDENTIAL_CAPTURE_KEY };
```
ezhang6811: do we need this?

lambda-layer/packages/layer/scripts/otel-instrument
```diff
@@ -39,7 +39,44 @@ _get_node_runtime_version() {
 
 NODE_VERSION=$(_get_node_runtime_version)
 
-if _is_esm_handler || [[ ${HANDLER_IS_ESM} == true ]] || [[ $NODE_VERSION -ge 24 ]]; then
+# Profile require() times when NODE_PROFILE_IMPORT_TIME=1 (analogous to PYTHONPROFILEIMPORTTIME=1)
+if [[ "${NODE_PROFILE_IMPORT_TIME}" == "1" ]]; then
+  export NODE_OPTIONS="${NODE_OPTIONS} --require /opt/profile-imports.js"
```
ezhang6811: just for testing/dissecting cold start, let's remove.

lambda-layer/packages/layer/scripts/otel-instrument
```diff
@@ -39,7 +39,38 @@ _get_node_runtime_version() {
...
+    export OTEL_METRICS_EXPORTER="none";
+fi
+
+# - Disable logs exporter by default
+if [ -z "${OTEL_LOGS_EXPORTER}" ]; then
+    export OTEL_LOGS_EXPORTER="none";
+fi
+
+# - Lite Mode defaults off; explicit OTEL_AWS_LAMBDA_FAST_START=true opts in.
+if [ -z "${OTEL_AWS_LAMBDA_FAST_START}" ]; then
```
wangzlei: nit: can remove these 3 lines

aws-distro-opentelemetry-node-autoinstrumentation/src/opentelemetry-lite-sdk.ts
```diff
@@ -0,0 +1,1112 @@
...
+function getHrTime(): HrTime {
+  const hr = process.hrtime();
+  const wallMs = _hrOffset + hr[0] * 1000 + hr[1] / 1e6;
+  const seconds = Math.floor(wallMs / 1000);
+  const nanos = Math.round((wallMs - seconds * 1000) * 1e6);
+  return [seconds, nanos];
+}
+
+function hrTimeToNanos(hrTime: HrTime): number {
+  return hrTime[0] * 1e9 + hrTime[1];
```
ezhang6811: use BigInt here so we don't lose precision when nanosecond value exceeds the max integer.

aws-distro-opentelemetry-node-autoinstrumentation/src/opentelemetry-lite-sdk.ts
```diff
@@ -0,0 +1,1112 @@
...
+// Always writes the field (no zero-skip), since the only caller — the span
+// `flags` field — always has the HAS_IS_REMOTE bit (0x100) set, so it is never
+// zero and must always be emitted.
+function encodeFixed32Field(fieldNumber: number, value: number): Buffer {
+  const buf = Buffer.alloc(4);
+  buf.writeUInt32LE(value >>> 0);
+  return Buffer.concat([encodeTag(fieldNumber, WIRE_FIXED32), buf]);
+}
+
+function encodeAnyValue(value: AttributeValue): Buffer {
```
ezhang6811, line 601: add a fallback for arrays here, same as you did in Python.

aws-distro-opentelemetry-node-autoinstrumentation/src/opentelemetry-lite-sdk.ts
```diff
@@ -0,0 +1,1112 @@
...
+      if (value !== undefined) this._attributes[key] = value;
+    }
+    return this;
+  }
+
+  addEvent(name: string, attributesOrStartTime?: Attributes | TimeInput, startTime?: TimeInput): this {
+    if (this._endTime !== undefined) return this;
+    let attrs: Attributes = {};
+    let time: TimeInput | undefined = startTime;
+    if (attributesOrStartTime && typeof attributesOrStartTime === 'object' && !Array.isArray(attributesOrStartTime)) {
```
ezhang6811: Confused by this logic. Are we saying definitively that attributesOrStartTime is a date if its object type is 'object' or an array? What about if its type is "Date"?

wespil: This version was meant to reject Date Objects by checking if attributesOrStartTime was an object but this type checking was too broad. 

Fix: include `instanceof Date` so that Date objects can be treated as timestamp rather than included in attributes

wangzlei (approved):

PR #501, opened 2026-07-07
refactor(serviceevents): framework-agnostic span processor + incident correlation/dedup hardening

vastin:
## Description

Refines the ServiceEvents endpoint span processor and incident-snapshot pipeline for the Node.js distro. Builds on the initial ServiceEvents feature (#467). Mirrors the Python distro changes.

Highlights:
- **Framework-agnostic span processor** — the endpoint span processor is now the sole endpoint path and framework-agnostic; the legacy per-framework hooks are removed. Renamed to `ServiceEventsSpanProcessor` to align with the Java distro.
- **Incident trace correlation gated on SAMPLED** — an incident snapshot carries `trace_id`/`span_id` only when the request's trace was actually sampled.
- **Incident flush interval derived from `maxSameError`** — `max(10s, 60s / N)`, maximizing the in-batch sampled-upgrade window.
- **Hardened incident dedup/rate-limiting** — the dedup hash recovers the error identity (type + message) from investigation data instead of collapsing to route-only.
- **Keep-alive operation leak fixed** — `onStart` force-clears the operation ALS before stamping the raw URL path, so a boundary span on a reused keep-alive socket no longer inherits the previous request's operation on its function-duration data points.

## Testing

- `tsc --noEmit` clean; ServiceEvents suite 463 passing; ESLint 0 errors, Prettier clean.
- Validated end-to-end against live CloudWatch on EC2 (5/5 ServiceEvents validators: deployment-event, incident-snapshot, incident-snapshot-latency, endpoint-error-metric, function-call-metric).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor / code hardening

## Checklist

- [x] Unit tests added/updated and passing
- [x] `CHANGELOG.md` updated


aws-distro-opentelemetry-node-autoinstrumentation/src/serviceevents/collectors/incident-snapshot-collector.ts
```diff
@@ -311,16 +366,48 @@ export class IncidentSnapshotCollector extends BaseCollector {
...
+  /**
+   * Hash the dedup key from route + recovered exception type/message.
+   *
+   * Keyed `route:<route>` for latency (no exception) or `route:<route>|exc:<type>:<message>` for
+   * errors, matching the documented per-SDK key (see SERVICE_EVENTS_INCIDENT_RATE_LIMITING.md) and
+   * the Python distro's `_error_hash`.
+   */
+  private errorHash(route: string, excType: string | null, excMessage: string): string {
+    const hashInput = excType === null ? `route:${route}` : `route:${route}|exc:${excType}:${excMessage}`;
     return crypto.createHash('md5').update(hashInput, 'utf-8').digest('hex');
```
jj22ee, line 411: The dedup hash is `route:<route>|exc:<type>:<message>` (line 410). Should this hash on the throw-site `file:line:col` instead of `exception.message`, in case the message contains a per-request variable (e.g. `"user 12345 not found"`)? A varying message makes every occurrence of the same logical error hash differently, so dedup never matches and the only backstop is the rate limiter (`maxPerMinute`, default 100) — or is hashing on `message` intended even when it embeds variables?

If `file:line:col` is preferable, the pieces are already here: `invData.exception` carries `traceback`, and `extractFunctionFromStackTrace` (endpoint-span-processor.ts:102) already parses the top `at` frame — the regex could also capture the `(/path/file.js:42:17)` location. (Note this mirrors Python #812's `_error_hash`, so any change should land in both.)

jj22ee (approved):

aws-distro-opentelemetry-node-autoinstrumentation/src/serviceevents/collectors/incident-snapshot-collector.ts
vastin: Good catch. It is a legacy issue of dedup logic before span processor refactoring. I will send follow-up PR to address it.

srprash (approved):

PR #502, opened 2026-07-08
feat: support lite mode on ESM handlers and Node 24

wespil:
## Description

Adds lite SDK (Fast Start) support for **ESM handlers** and **Node.js 24**.

Lite mode instruments via `require-in-the-middle` and prototype patching only.
That works for CommonJS handlers, but `require-in-the-middle` cannot intercept
ES modules — so on ESM handlers, and on Node 24 (where the runtime loads modules
as ESM), lite mode produced **no spans**. This PR adds an ESM entry point and
routes to it when the CJS fast path can't work.

## Changes

- **`src/lite-wrapper.mjs`** (new) — ESM entry point that installs the
  `import-in-the-middle` loader hook before running `configureLiteMode()`,
  mirroring `wrapper.mjs` for the full SDK.
- **`scripts/otel-instrument`** — the lite-mode branch now selects the wrapper
  by handler type / Node version:
  | Handler / Node | Routing |
  |---|---|
  | CJS, Node < 24 | `--require /opt/lite-wrapper.js` (unchanged fast path) |
  | ESM, Node 20+ | `--import /opt/lite-wrapper.mjs` |
  | ESM, Node 18 | `--import /opt/lite-wrapper.js --experimental-loader=import-in-the-middle/hook.mjs` |
  | Node 24 (any) | `--import /opt/lite-wrapper.mjs` |
- **`package.json`** — copy `lite-wrapper.mjs` into the layer on `postcompile`;
  pin `typescript` to `^5.0.4` so the lockless build install can't resolve to
  TypeScript 7 (which breaks `ts-loader`).

## Trade-off

Loading `import-in-the-middle` on ESM / Node 24 reintroduces the ESM hook cost
that lite mode is designed to avoid, so the cold-start advantage is smaller on
those runtimes. The fast, hook-free path is preserved for CJS on Node < 24.
Correctness (spans are actually produced) takes priority over the optimization.

## Testing
- NodeJS Main Build PASS https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/28968338850


.github/workflows/application-signals-e2e-test.yml
```diff
@@ -131,7 +131,14 @@ jobs:
       aws-region: us-east-1
       caller-workflow-name: 'main-build'
 
-  
+  lambda-lite-sdk:
+    uses: aws-observability/aws-application-signals-test-framework/.github/workflows/node-lambda-lite-sdk-test.yml@otel-lite-sdk
```
wespil: Will update to Main for merge

ezhang6811 (dismissed):

wangzlei (approved):

PR #506, opened 2026-07-16
Fix import path for BatchSpanProcessorBase

PawelDmochowskiMeetingPackage:
Version 2.9.0 of `@opentelemetry/sdk-trace-base` no longer contains this export as it has been refactored to `@opentelemetry/sdk-trace`

*Issue #, if available:*

*Description of changes:*


srprash:
Looks like this Nightly build PR for upstream bump already includes the fix through a shim: https://github.com/aws-observability/aws-otel-js-instrumentation/pull/499/files#r3861288096

srprash:
I think the approach in nightly build PR (499) is better since it gets rid of the deep import. 
Thanks for surfacing the change but closing this PR in favor of 499.

PR #520, opened 2026-08-12
Add CloudWatch plugin for OpenTelemetry JS (span metrics)

srprash:
## Description

Adds the `cloudwatch-plugin-otel` workspace — an OpenTelemetry JS extension that generates span-derived RED metrics (`traces.span.metrics.calls`, `traces.span.metrics.duration`) **in-process, from 100% of spans, before trace sampling**, so metrics stay accurate even at low trace-sampling rates (unlike deriving them downstream from an already-sampled stream).

### How it works
- **`SpanMetricsProcessor`** records the two metrics in `onEnd()` via the host SDK's `MeterProvider`.
- **`AlwaysRecordSampler`** wraps the configured sampler (`DROP` → `RECORD_ONLY`) so unsampled spans are still recorded (and metered) but not exported — trace volume/cost stay at the configured rate.

### Integration modes
- **Programmatic** — `new NodeSDK(withSpanMetrics(config))`
- **Manual** — `SpanMetricsProcessor` + `AlwaysRecordSampler` + `bind()`
- **Zero-code** — `node --require @aws/cloudwatch-plugin-otel/register ...` (patches the `NodeSDK` layout detected at runtime; self-verifies after `start()` and disables safely on an unrecognized layout)

### Metric/attribute shape
Follows the SpanMetricsConnector schema and OTel semantic conventions: short-form base dimensions (`service.name`, `span.name`, `span.kind`, `status.code`), an allowlisted low-cardinality subset of derived semconv attributes per family, DB legacy-key passthrough, and messaging temporary/anonymous-destination omission. Instrumentation scope name: `cloudwatch.plugin.otel.span_metrics`.

## Testing
- **Unit tests** — sampler, processor emission, attribute builder (per-family allowlist, legacy fallback, omissions), holder, shape detection, exception-safety.
- **Contract tests** — mode wiring (all modes: 100% metrics vs sampled trace export) and per-family attribute shape (base, HTTP, RPC, DB, messaging) driven through **real instrumentation** to a mock OTLP collector. DB (Postgres + `pg`) and messaging (Kafka + `kafkajs`) use Testcontainers; RPC (gRPC) runs in-process; both skip gracefully when Docker is unavailable.
- Verified across `@opentelemetry/sdk-node` 0.203–0.221 (both NodeSDK internal layouts) and against the latest published OpenTelemetry packages.

## CI
Adds a path-scoped PR build workflow (`cloudwatch-plugin-otel-pr-build.yml`): lint, unit tests, and a unit+contract matrix against pinned and latest OpenTelemetry packages.

## Notes
Opened as a draft. Follow-ups before marking ready: `CHANGELOG.md` and a release/publish workflow to match the other packages.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


cloudwatch-plugin-otel/src/span-metrics.ts
```diff
@@ -0,0 +1,72 @@
...
+ * the user's span export. We mirror the SDK's own default wrapping (new BatchSpanProcessor(exporter))
+ * and remove traceExporter so the SDK does not see both.
+ *
+ * The deprecated singular spanProcessor option is not supported: it is warned about and skipped.
+ *
+ * Pure transform on public config fields; no patching. The caller still binds the MeterProvider
+ * after start via {@link bind} (the SDK owns it).
+ */
+export function withSpanMetrics<T extends NodeSdkConfigLike>(config: T): T {
+  if (config.sampler) {
```
liustve: Blocking: `withSpanMetrics()` wraps only an explicit `config.sampler`. When the recommended Mode 2 example leaves that field unset and sampling is configured through `OTEL_TRACES_SAMPLER`, NodeSDK resolves the environment sampler later and the wrapper never sees it. With `always_off`, I reproduce 0 metrics from 100 spans here versus 100/100 when the same sampler is explicit. That violates the extension's core 100%-metrics contract. Please resolve and wrap the environment sampler with the host SDK's exact validation semantics, and add a Mode 2 contract case with no explicit sampler.

cloudwatch-plugin-otel/src/span-metrics-processor.ts
```diff
@@ -0,0 +1,84 @@
...
+  // Initialized on the first onEnd() once the host has supplied a MeterProvider (late-bind).
+  private calls: Counter | undefined;
+  private duration: Histogram | undefined;
+
+  onStart(span: Span, _parentContext: Context): void {
+    // Dedup + schema/lib markers: presence of the schema attribute tells the backend this span's
+    // metrics were already generated by the extension. Guarded so a failure here can never
+    // propagate into the host application's span pipeline.
+    try {
+      span.setAttribute(SCHEMA_ATTR, SCHEMA_VERSION);
```
liustve, line 48: Blocking: this stamps the backend dedup marker before knowing whether this processor can generate a metric. `onEnd()` returns when no MeterProvider is bound, so with `OTEL_METRICS_EXPORTER=none` an exported span still carries `aws.otel.span.metrics.schema=v1` even though the client recorded no metric. The backend then skips generation too. Please stamp only when a usable provider is bound and add a regression test asserting both no metric and no marker in the unbound case. Moving this to `onEnd()` is not sufficient because ended spans reject attribute mutation.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+          let processors: any[];
+          if (cfg.spanProcessors) {
+            processors = [...cfg.spanProcessors];
+          } else if (cfg.spanProcessor) {
+            processors = [cfg.spanProcessor];
+            delete cfg.spanProcessor;
+          } else {
+            // traceExporter / env: let the SDK's own env resolution stand, then add ours. (A lone
+            // traceExporter is rare for register users, who are env-driven; env is the common path.)
+            processors = envSpanProcessors();
```
liustve: Blocking: this branch also handles an explicit `cfg.traceExporter`, but it replaces that exporter with environment processors and then assigns `cfg.spanProcessors`. On the CONFIG layout, upstream's precedence causes `traceExporter` to be ignored. I reproduce a custom exporter being removed entirely when `OTEL_TRACES_EXPORTER=none`. Please preserve it by applying the same `BatchSpanProcessor(traceExporter)` conversion used by `withSpanMetrics()`, and add a register-mode traceExporter contract case.

cloudwatch-plugin-otel/src/internal/detect-shape.ts
```diff
@@ -0,0 +1,31 @@
...
+// UNKNOWN means neither was recognized on a probe instance -> the caller disables the patch.
+export type SdkShape = 'FIELD' | 'CONFIG' | 'UNKNOWN';
+
+// Construct a throwaway NodeSDK and inspect which internal field it uses to hold span-processor
+// config. Detecting the layout (rather than matching a version number) self-adapts to new sdk-node
+// releases that keep a known shape, and fails safe on ones that don't. Any error, or neither field
+// being present, is UNKNOWN.
+export function detectShape(NodeSDK: new (config: Record<string, unknown>) => unknown, version: string): SdkShape {
+  try {
+    const probe = new NodeSDK({ spanProcessors: [] }) as Record<string, unknown>;
```
liustve: Blocking: constructing this throwaway NodeSDK is not side-effect-free. With `OTEL_METRICS_EXPORTER=prometheus`, its constructor eagerly creates a PrometheusExporter that binds port 9464. The real SDK then loses the port, while the orphan still returns HTTP 200 with "MetricReader is not bound to a MetricProducer"; application and RED metrics are both absent. Please make the probe side-effect-free (including empty metric/log processor arrays and disabled resource detection), or detect without constructing an SDK, and cover Prometheus on both runtime layouts.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+// Resolve the sampler the SDK would build from OTEL_TRACES_SAMPLER(+_ARG), using the same
+// sdk-trace-base copy the SDK uses, so our wrapper delegates to the user's configured sampler.
+// Mirrors the OTel env sampler mapping; defaults to ParentBased(AlwaysOn) like the SDK.
+function resolveEnvSampler(sdkNodePath: string): any {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const base = require(require.resolve('@opentelemetry/sdk-trace-base', { paths: [sdkNodePath] }));
+  const { ParentBasedSampler, AlwaysOnSampler, AlwaysOffSampler, TraceIdRatioBasedSampler } = base;
+  const name = process.env.OTEL_TRACES_SAMPLER;
+  const argRaw = process.env.OTEL_TRACES_SAMPLER_ARG;
+  const ratio = argRaw !== undefined && argRaw !== '' ? Number(argRaw) : 1;
```
liustve: This parser diverges from upstream for malformed or out-of-range `OTEL_TRACES_SAMPLER_ARG`. Values such as `abc`, `-1`, or whitespace become ratio 0 here, while the host SDK warns and falls back to ratio 1. The plugin therefore silently removes all trace export while its 100% metrics mask the failure. Please reuse or exactly match the host resolver and test invalid, blank, negative, and greater-than-one inputs.

cloudwatch-plugin-otel/contract-tests/e2e/mode.test.ts
```diff
@@ -0,0 +1,106 @@
...
+      });
+
+      it('counts 100% of spans while traces are sampled below the request count', () => {
+        assert.strictEqual(
+          collector.callsValue(SERVER_SPAN_NAME),
+          REQUEST_COUNT,
+          `calls for ${SERVER_SPAN_NAME} should equal ${REQUEST_COUNT}`
+        );
+        assert.ok(
+          collector.countExportedSpans(SERVER_SPAN_NAME) < REQUEST_COUNT,
```
liustve: This assertion accepts zero exported spans, so complete trace-export loss remains green. That is exactly what the CONFIG traceExporter regression does. Please assert trace preservation deterministically, for example with an always-on/always-off pair, or otherwise require a stable two-sided bound without introducing probabilistic flakes.

.github/workflows/cloudwatch-plugin-otel-pr-build.yml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Build and run unit tests
+        run: npm test
+
+  compat-tests:
+    name: Unit + contract (OTel ${{ matrix.channel }})
+    runs-on: ubuntu-latest # provides a Docker daemon for the DB/messaging Testcontainers suites
+    strategy:
+      fail-fast: false
+      matrix:
+        channel: [pinned, latest] # pinned = the versions the module declares; latest = newest release
```
liustve: The compatibility matrix currently has no meaningful version boundary: the declared pinned range and `latest` both resolve to sdk-node 0.221, so neither job exercises the FIELD injection path used by 17 of the 19 versions advertised in the README. The README also cites a scheduled "Span Metrics Extension - Compatibility" workflow that is not present. Please add a real lower-bound version axis and scheduled run, or narrow the compatibility claims to what CI preserves.

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+```js
+const { NodeSDK } = require('@opentelemetry/sdk-node');
+const { getNodeAutoInstrumentations } = require('@opentelemetry/auto-instrumentations-node');
+const { MeterProvider, PeriodicExportingMetricReader } = require('@opentelemetry/sdk-metrics');
+const { OTLPMetricExporter } = require('@opentelemetry/exporter-metrics-otlp-http');
+const { withSpanMetrics, bind } = require('@aws/cloudwatch-plugin-otel');
+
+// The extension records into a MeterProvider you own; construct and bind it.
+const meterProvider = new MeterProvider({
```
liustve: The recommended Mode 2 recipe constructs a separate MeterProvider without the NodeSDK resource and never installs it globally. Span metrics therefore export under a default `unknown_service:node` resource, while NodeSDK creates no metric provider because it receives no readers; all other instrumentation metrics remain on the no-op global provider. Please replace this with one correct single-provider recipe, bind the SDK-owned global provider only after `start()`, and execute the example in a test. The no-op first-bind latch must be fixed with this change.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+  const OriginalNodeSDK = sdkNode.NodeSDK;
+
+  // Detect the internal shape by probing a throwaway instance rather than matching a version number.
+  // A version allowlist both rejects new-but-compatible versions and accepts a version whose
+  // internals silently shifted; probing the actual layout self-adapts and is verified again after
+  // start(). The probe passes a minimal trace config so the FIELD-shape constructor populates
+  // _tracerProviderConfig; CONFIG-shape leaves that undefined and exposes _configuration instead.
+  const shape = detectShape(OriginalNodeSDK, version);
+  if (shape === 'UNKNOWN') {
+    diag.warn(
```
liustve, line 88: These preload-time diagnostics are invisible under the default configuration because the upstream register has not installed a DiagLogger yet. I reproduce the UNKNOWN-layout path printing nothing unless `OTEL_LOG_LEVEL` is set, contradicting the "warns loudly" fail-safe claim. Please use a warning path guaranteed to be visible during preload without permanently replacing the application's logger.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+
+    start(): void {
+      try {
+        const self = this as Record<string, any>;
+        if (isConfigShape) {
+          // Config shape (>= 0.220): start() derives spanProcessors from this._configuration
+          // (spanProcessors ?? [spanProcessor] ?? [batch(traceExporter)] ?? env). Normalize that to
+          // an explicit spanProcessors list with ours appended, so exporting is preserved.
+          const cfg = self._configuration ?? (self._configuration = {});
```
liustve, line 186: nit: `_configuration` is the caller's object by reference, so these writes mutate reusable config; duplicate preload copies can also append this processor twice. Can we shallow-copy before mutation and place an idempotence marker on the sdk-node export object?

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+
+  // Env span processors the SDK would build when the user configured none (zero-code). Resolved from
+  // the same sdk-node copy the SDK uses, so exporting keeps working after we add our processor.
+  function envSpanProcessors(): any[] {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const utils = require(require.resolve('@opentelemetry/sdk-node/build/src/utils', { paths: [sdkNodePath] }));
+      return utils.getSpanProcessorsFromEnv ? utils.getSpanProcessorsFromEnv() : [];
+    } catch {
+      return [];
```
liustve: nit: on any private-helper resolution failure this returns `[]`, after which the CONFIG path replaces upstream exporters with only `SpanMetricsProcessor`. The path resolves today, but the private signature has already changed. Can failure leave upstream's own env fallback intact instead of turning a future helper change into silent trace loss?

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+  // upstream will construct by resolving sdk-node relative to auto-instrumentations-node.
+  let sdkNodePath: string;
+  let pkgPath: string;
+  try {
+    const autoInstrEntry = require.resolve('@opentelemetry/auto-instrumentations-node');
+    sdkNodePath = require.resolve('@opentelemetry/sdk-node', { paths: [autoInstrEntry] });
+    pkgPath = require.resolve('@opentelemetry/sdk-node/package.json', { paths: [autoInstrEntry] });
+  } catch {
+    // Fall back to our own resolution (e.g. manual/programmatic users without auto-instrumentations).
+    sdkNodePath = require.resolve('@opentelemetry/sdk-node');
```
liustve, line 73: nit: if both resolution attempts fail, the fallback throw escapes module evaluation and `--require @aws/cloudwatch-plugin-otel/register` terminates the application. Can we guard the fallback/top-level patch and emit one direct prerequisite error?

cloudwatch-plugin-otel/src/span-metrics-processor.ts
```diff
@@ -0,0 +1,84 @@
...
+      this.calls!.add(1, attributes);
+      // hrtime duration [seconds, nanos] -> seconds as a double (spec §1).
+      const [seconds, nanos] = span.duration;
+      this.duration!.record(seconds + nanos / 1e9, attributes);
+    } catch (e) {
+      diag.warn('Failed to record span metrics', e);
+    }
+  }
+
+  forceFlush(): Promise<void> {
```
liustve, line 70: nit: these no-op lifecycle methods are appropriate because the processor does not own the MeterProvider, but spans ending after SDK shutdown begins lose metrics silently. Can the README state the required order: stop intake, wait for in-flight spans, then shut down the SDK?

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,141 @@
...
+  readers: [new PeriodicExportingMetricReader({ exporter: new OTLPMetricExporter() })],
+});
+bind(meterProvider);
+
+const sdk = new NodeSDK(withSpanMetrics({
+  traceExporter: new OTLPTraceExporter(),
```
liustve, line 52: nit: this programmatic example uses `OTLPTraceExporter` without importing it, so it fails as written. Can we add the missing exporter import and execute the example as a documentation test?

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+`withSpanMetrics(config)` wraps your sampler (if set), appends the span processor, and — if you
+configured only `traceExporter` — converts it into a `BatchSpanProcessor` so your span export is
+preserved (NodeSDK ignores `traceExporter` once `spanProcessors` is set). The deprecated singular
+`spanProcessor` option is not supported (a warning is logged and it is ignored).
+
+### Manual
+
+```js
+const { TracerProvider, BatchSpanProcessor } = require('@opentelemetry/sdk-trace-node');
```
liustve: nit: `@opentelemetry/sdk-trace-node` 2.x does not export `TracerProvider`; `NodeTracerProvider` is the public class, so this manual example fails as written. Can we use `NodeTracerProvider` and execute the example as a documentation test?

cloudwatch-plugin-otel/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+The extension records only into your existing metrics pipeline (it adds a metric *source*, never a
+new exporter or route) and keeps your sampling configuration unchanged.
+
+The table below reflects results **verified by the extension's unit + contract suites run
+end-to-end** (all modes, real instrumentation → OTLP → collector). A nightly
+`Span Metrics Extension - Compatibility` workflow re-runs them against the latest published
+OpenTelemetry packages so drift shows up as a red run.
+
+All three modes are verified end-to-end across the full range **`@opentelemetry/sdk-node` 0.203 →
```
liustve, line 142: nit: the verified dependency pairings are inaccurate. sdk-node 0.203 uses sdk-trace-base/sdk-metrics 2.0.1, 0.219 uses 2.8.0, 0.220 uses 2.9.0, and 0.221 uses 2.10.0. The stated 2.7 floor also conflicts with this package's own `^2.0.0` dependency range. Can we update the table to the tested versions?

cloudwatch-plugin-otel/test/detect-shape.test.ts
```diff
@@ -0,0 +1,46 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import * as assert from 'assert';
+import { detectShape } from '../src/internal/detect-shape';
+
+// Fakes standing in for the two known NodeSDK layouts and the failure cases. The register patch
+// relies on detectShape to choose an injection strategy and to disable safely on anything else.
+class FieldShapeSdk {
```
liustve, line 12: nit: real FIELD-layout NodeSDK instances also contain `_configuration`; the FIELD result depends on checking `_tracerProviderConfig` first. These fakes each expose only one field, so reversing the production branch order would break 17 supported versions while this test stayed green. Can we add a both-fields-present case that must resolve to FIELD?

cloudwatch-plugin-otel/src/internal/open-telemetry-holder.ts
```diff
@@ -0,0 +1,31 @@
...
+
+// Bridges the host's fully-built MeterProvider to the span processor. The processor is constructed
+// before the MeterProvider exists, so the provider is supplied later by whichever host hook fires:
+// the register patch, withSpanMetrics(), or a manual bind() call. Internal, not public API.
+
+let instance: MeterProvider | undefined;
+
+// First bind wins. A later bind of a different provider is ignored but logged.
+export function set(meterProvider: MeterProvider): void {
+  if (instance === undefined) {
```
liustve, line 26: nit: a pre-start `metrics.getMeterProvider()` is a NoopMeterProvider, but first-bind-wins makes that binding permanent and the processor then caches no-op instruments. Since the corrected Mode 2 recipe will bind the SDK global provider after `start()`, can the holder reject/lazily replace a no-op provider?

liustve, line 26: nit: after the first provider is shut down, this holder can never bind a replacement, so an in-process SDK restart records zero RED metrics. Can we either support a shutdown-aware rebind or explicitly document that restart in one process is unsupported?

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,222 @@
...
+
+function patch(): void {
+  // auto-instrumentations-node may resolve its OWN nested copy of sdk-node, which is a different
+  // module instance than a bare require('@opentelemetry/sdk-node') from here. Patch the SAME copy
+  // upstream will construct by resolving sdk-node relative to auto-instrumentations-node.
+  let sdkNodePath: string;
+  let pkgPath: string;
+  try {
+    const autoInstrEntry = require.resolve('@opentelemetry/auto-instrumentations-node');
+    sdkNodePath = require.resolve('@opentelemetry/sdk-node', { paths: [autoInstrEntry] });
```
liustve, line 69: nit: this patches only the sdk-node copy resolved by auto-instrumentations-node. A different copy constructed by the app, and the in-repo `OTEL_AWS_LAMBDA_FAST_START=true` lite-SDK path, bypass the patch and self-check silently. Can we document these zero-code limitations?

.github/workflows/release-cloudwatch-plugin-otel.yml
```diff
@@ -0,0 +1,61 @@
...
+jobs:
+  build:
+    environment: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Contrib Repo @ SHA - ${{ github.sha }}
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
+
+      - name: Set up Node and run Unit Tests
+        uses: ./.github/actions/set_up
```
liustve, line 24: nit: this reused action runs `npm install` rather than `npm ci`, ignores its `package_name` input, and compiles/tests every workspace. Can we make the release checks reproducible and scoped to `@aws/cloudwatch-plugin-otel`?

cloudwatch-plugin-otel/tsconfig.json
```diff
@@ -0,0 +1,37 @@
+{
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "skipLibCheck": true,
+
+    "allowUnreachableCode": false,
+    "allowUnusedLabels": false,
+    "declaration": true,
+    "declarationMap": true,
```
liustve, line 10: nit: `declarationMap` emits `.d.ts.map` files, but the package's `files` allowlist ships only `.js`, `.js.map`, and `.d.ts`. Can we either include the declaration maps in the published artifact or stop generating them?

cloudwatch-plugin-otel/package.json
```diff
@@ -0,0 +1,97 @@
...
+  "version": "0.0.1",
+  "description": "Generates span-derived RED metrics inside the OpenTelemetry JS SDK, from 100% of spans, before trace sampling.",
+  "author": {
+    "name": "Amazon Web Services",
+    "url": "http://aws.amazon.com"
+  },
+  "homepage": "https://github.com/aws-observability/aws-otel-js-instrumentation/tree/main/cloudwatch-plugin-otel#readme",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=18"
```
liustve, line 12: nit: this advertises all Node 18 releases, but the OTel 2.x dependencies require `^18.19.0 || >=20.6.0`. Can we align the package engine range with its runtime dependencies?

.github/workflows/release-cloudwatch-plugin-otel.yml
```diff
@@ -0,0 +1,61 @@
...
+          fi
+
+      - name: Upgrade npm
+        run: npm install -g npm@latest
+
+      - name: Publish to npm
+        working-directory: cloudwatch-plugin-otel
+        env:
+          NPM_CONFIG_PROVENANCE: true
+        run: npm publish
```
liustve, line 71: Question: this publish has no `NODE_AUTH_TOKEN` and appears to rely entirely on npm OIDC trusted publishing. How will the first release of this new scoped package be bootstrapped, given that trusted publisher configuration normally attaches to an existing package?

cloudwatch-plugin-otel/src/span-metrics-processor.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  private ensureInstruments(): boolean {
+    if (this.calls !== undefined) {
+      return true;
+    }
+    const meterProvider = holder.get();
+    if (meterProvider === undefined) {
+      return false;
+    }
+    const meter: Meter = meterProvider.getMeter(SCOPE_NAME);
```
liustve, line 86: Spec question: §5 marks the instrumentation-scope row OPEN and gives only a Java example. What scope name should JavaScript use, and is the library version required as the scope version? The current code uses `cloudwatch.plugin.otel.span_metrics` without a version; I am not treating either choice as a defect until the OPEN row is resolved.

cloudwatch-plugin-otel/src/span-metrics-attributes-builder.ts
```diff
@@ -0,0 +1,106 @@
...
+import { Attributes, SpanKind, SpanStatusCode } from '@opentelemetry/api';
+import { ReadableSpan } from '@opentelemetry/sdk-trace-base';
+import { LIB_VERSION, LIB_VERSION_ATTR, SCHEMA_ATTR, SCHEMA_VERSION } from './identity';
+
+const SERVICE_NAME = 'service.name';
+
+// Copied when present, regardless of span family (a span only carries the keys of its own family,
+// so no family branching is needed). Value TYPES are preserved as the span carries them:
+// http.response.status_code stays a number per semconv; everything else is a string.
+const ALLOWLIST: string[] = [
```
liustve, line 11: Spec question: §4's closing prose says `rpc.grpc.status_code` is an integer, but the RPC table omits it and the flat-allowlist instruction points implementers to the table. Should this key be included in v1? The current code follows the table.

cloudwatch-plugin-otel/src/span-metrics-attributes-builder.ts
```diff
@@ -0,0 +1,106 @@
...
+// Builds the metric attribute set: four base dimensions, any allowlisted semconv attribute present
+// on the span, and the identity/schema markers.
+export function buildAttributes(span: ReadableSpan): Attributes {
+  const attributes: Attributes = {
+    'span.name': span.name,
+    'span.kind': SPAN_KIND_NAMES[span.kind] ?? 'INTERNAL',
+    'status.code': STATUS_CODE_NAMES[span.status.code] ?? 'UNSET',
+    // Schema + library-version markers appear on both spans and metrics (spec §6).
+    [SCHEMA_ATTR]: SCHEMA_VERSION,
+    [LIB_VERSION_ATTR]: LIB_VERSION,
```
liustve, line 68: Spec question: §6 currently requires `aws.otel.extension.lib.version` on spans and generated metrics but says it will likely move to a resource attribute. Should v1 remain as written, or should this PR adopt the resource placement now?

cloudwatch-plugin-otel/src/span-metrics.ts
srprash: Fixed in eddf096. `withSpanMetrics` now resolves the env sampler when `config.sampler` is unset and always wraps a concrete sampler (falling back to the SDK default), so record-forcing applies when sampling is configured via OTEL_TRACES_SAMPLER. Added a unit test plus an always_off end-to-end contract test.

cloudwatch-plugin-otel/src/span-metrics-processor.ts
srprash: Fixed in 96f706a. The dedup marker is now stamped in onStart only when a MeterProvider is bound (holder.hasProvider()); with metrics disabled the SDK builds no provider, so the span is not marked and the backend still generates the metric. Marking stays in onStart because the span is immutable by onEnd.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 97a7ff5. The CONFIG path now preserves an explicit traceExporter by wrapping it into a BatchSpanProcessor (mirroring NodeSDK) and appending ours, instead of dropping it. Extracted into a pure normalize-span-processors module with a unit test covering the traceExporter case; the contract export assertion was tightened to catch zero-export loss.

cloudwatch-plugin-otel/src/internal/detect-shape.ts
srprash: Fixed in 5c83a1f. detectShape no longer constructs a NodeSDK; it inspects the constructor source (NodeSDK.toString()) for the _tracerProviderConfig field, so there are no construction side effects (no Prometheus server, etc.). UNKNOWN fail-safe retained. Verified against real 0.203 (FIELD) and 0.221 (CONFIG).

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in eddf096. Replaced the hand-rolled parser with a shared resolver built on core's getNumberFromEnv and public sampler classes. Blank/non-numeric/negative/>1 args now fall back to ratio 1 (matching upstream), never 0. Edge cases covered by unit tests.

cloudwatch-plugin-otel/contract-tests/e2e/mode.test.ts
srprash: Fixed in 97a7ff5. The assertion is now two-sided (0 < exported < REQUEST_COUNT) with 50% sampling for a deterministic floor, so complete export loss fails. Also added an always_off case asserting 100% metrics with zero exported.

.github/workflows/cloudwatch-plugin-otel-pr-build.yml
srprash: Fixed in 5c83a1f. Added a field-floor matrix leg pinning sdk-node 0.203.0 (with auto-instrumentations 0.62.0 and the 2.0.1 SDK packages), which exercises the FIELD injection path that pinned/latest (both CONFIG) never hit. Unit + contract suites pass against 0.203. The nonexistent nightly-workflow reference was removed from the README.

cloudwatch-plugin-otel/README.md
srprash: Fixed in 5c83a1f. The programmatic recipe no longer builds a detached MeterProvider. NodeSDK registers its MeterProvider globally, so the processor now resolves that provider automatically; metrics inherit the SDK resource (service.name) instead of unknown_service:node. Verified end-to-end. bind() is retained for hosts that keep a non-global provider.

cloudwatch-plugin-otel/README.md
srprash: Fixed in 5c83a1f. The programmatic example now imports OTLPTraceExporter and was executed end-to-end.

cloudwatch-plugin-otel/README.md
srprash: Fixed in 5c83a1f. The manual example now uses NodeTracerProvider (and imports its dependencies); executed end-to-end.

cloudwatch-plugin-otel/README.md
srprash: Fixed in 5c83a1f. Corrected the pairings to 0.203->2.0.1, 0.219->2.8.0, 0.220->2.9.0, 0.221->2.10.0 and restated the range accordingly.

cloudwatch-plugin-otel/test/detect-shape.test.ts
srprash: Fixed in 5c83a1f. Added a fake that declares both fields and asserts FIELD, so the branch order (checking _tracerProviderConfig first) is now guarded; reversing it would fail. Also added a no-construction assertion.

cloudwatch-plugin-otel/src/internal/open-telemetry-holder.ts
srprash: Fixed in 5c83a1f. The holder now rejects a Noop provider (via a behavioral check) and resolves lazily, so a pre-start noop is never cached; instruments are created only once a usable provider is available. A single usable-provider predicate is shared by onStart and onEnd.

cloudwatch-plugin-otel/src/internal/env-sampler.ts
```diff
@@ -0,0 +1,103 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { diag } from '@opentelemetry/api';
+import { getNumberFromEnv, getStringFromEnv } from '@opentelemetry/core';
```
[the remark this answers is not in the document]
srprash: Fixed in 5592300. Inlined the two env reads (envString/envNumber) and removed the @opentelemetry/core import entirely, rather than adding a dependency — the package no longer requires core at runtime, so a standalone install cannot break on its absence or on a helper-signature mismatch. The inlined parsing matches core's getStringFromEnv/getNumberFromEnv behavior (blank/whitespace treated as unset; Number() parse with NaN as unset), verified differentially against core across the edge cases, with unit tests pinning that contract.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  Object.defineProperty(sdkNode, 'NodeSDK', {
+    enumerable: true,
+    configurable: true,
+    get: () => SpanMetricsNodeSDK,
+  });
+  diag.info('[span-metrics] NodeSDK patched');
+}
+
+patch();
```
liustve: **`register` can crash the host process at load when `@opentelemetry/sdk-node` isn't installed.**

`patch()` is invoked unguarded at the top level here. Its dependency-resolution `catch` fallback (lines 60–63) re-calls `require.resolve('@opentelemetry/sdk-node')`, which **throws out of `patch()`** with no outer guard. And `@opentelemetry/sdk-node` is a **`devDependency` only** — it is not in `dependencies` or `peerDependencies` in `package.json`, so a consumer of `@aws/cloudwatch-plugin-otel` does not get it transitively.

Net effect: a host-wide `NODE_OPTIONS="--require @aws/cloudwatch-plugin-otel/register"` will crash *any* Node process on the box that lacks `sdk-node` (a build step, a helper CLI, etc.) at startup. That contradicts this file's own stated contract (lines 11–14: "warns loudly and leaves telemetry running … rather than silently mis-wiring").

Repro: hide `@opentelemetry/sdk-node` via a `Module._resolveFilename` hook, then `require('.../register.js')` → throws `MODULE_NOT_FOUND` at load.

Suggested guard so a missing/absent SDK degrades to a no-op instead of a crash:

```suggestion
try {
  patch();
} catch (e) {
  diag.warn(
    '[span-metrics] initialization failed (is @opentelemetry/sdk-node installed?); ' +
      'span metrics are DISABLED. Traces/metrics continue with upstream behavior.',
    e
  );
}
```

cloudwatch-plugin-otel/src/span-metrics.ts
```diff
@@ -0,0 +1,77 @@
...
+export function withSpanMetrics<T extends NodeSdkConfigLike>(config: T): T {
+  // Force-record the configured sampler. The base is the explicit config.sampler, or — when the user
+  // configures sampling through OTEL_TRACES_SAMPLER(+_ARG) and leaves config.sampler unset — the env
+  // sampler, or the SDK's own default when neither is set. We always resolve a concrete sampler and
+  // wrap it (rather than leaving it unset for the SDK to build) so this mode force-records in exactly
+  // the same cases as zero-code, instead of relying on the SDK default happening to be AlwaysOn.
+  const base = config.sampler ?? resolveEnvSamplerOrDefault();
+  config.sampler = AlwaysRecordSampler.create(base);
+
+  if (config.spanProcessor) {
```
liustve: **`withSpanMetrics` silently drops a caller's deprecated singular `spanProcessor`, and disagrees with the register path on this.**

This branch only warns — it never converts or deletes `config.spanProcessor`. The value stays on the config while `config.spanProcessors` is set below; per NodeSDK's precedence (`spanProcessors ?? [spanProcessor] ?? …`), `spanProcessors` wins, so the caller's processor is **discarded** and its span export is lost — masked by the 100% metrics.

Repro: `withSpanMetrics({ spanProcessor: userProc })` → `config.spanProcessors = [SpanMetricsProcessor]`, `config.spanProcessor === userProc` (left dangling), `spanProcessors.includes(userProc) === false`.

I realize "singular `spanProcessor` is not supported" is a deliberate, documented choice (README + the docstring above). Two things worth deciding explicitly:

1. **Inconsistency:** the register CONFIG path *does* honor a singular `spanProcessor` (`src/internal/normalize-span-processors.ts:37-39` converts + deletes it), so the two supported wiring paths behave differently for the same input. Intentional?
2. **Silent loss:** if it stays unsupported, consider `delete config.spanProcessor` (so it doesn't dangle) and make the warning state that the processor's **span export will be dropped** — today a user who set only a singular processor loses all span export with just a generic "ignored" warning, which is exactly the metrics-mask-hides-trace-loss failure mode this package is otherwise careful about.

cloudwatch-plugin-otel/contract-tests/apps/programmatic-app.ts
```diff
@@ -0,0 +1,49 @@
...
+import { NodeSDK } from '@opentelemetry/sdk-node';
+import { BatchSpanProcessor, ParentBasedSampler, TraceIdRatioBasedSampler } from '@opentelemetry/sdk-trace-base';
+import { bind, withSpanMetrics } from '../../src';
+
+const endpoint = process.env.COLLECTOR_ENDPOINT ?? 'http://localhost:4319';
+const resource = resourceFromAttributes({ 'service.name': 'contract-programmatic' });
+
+// The app owns its metrics pipeline and binds it — the extension records into the app's own
+// MeterProvider rather than reaching into NodeSDK's internal one (whose config surface varies by
+// sdk-node version). This is the documented Mode 2 pattern.
```
liustve: **This comment mischaracterizes the recommended Mode 2 pattern, and the recommended pattern has no e2e coverage.**

The README's **recommended** Programmatic example uses `withSpanMetrics` with **no `bind()`** — it states "no bind() call is needed in this mode" and relies on the extension picking up the globally-registered MeterProvider (README "Programmatic (recommended)"). This app instead exercises the *optional* explicit-`bind()` variant (README: "Call `bind(provider)` only if you keep a `MeterProvider` you do **not** register globally").

So:
- This comment calling the explicit-bind approach "the documented Mode 2 pattern" is inaccurate — the documented/recommended pattern is the no-`bind()` one.
- No contract app covers the recommended path, i.e. the holder's global-provider fallback (`src/internal/open-telemetry-holder.ts:52-58`), which is currently unit-tested only. A future change to how NodeSDK registers the global MeterProvider could silently break the *recommended* path with green CI.

Suggestion: add (or repurpose) a contract app that uses `withSpanMetrics` **without** `bind()`, and reword this comment. (The recommended path does work today: after `start()`, `metrics.getMeterProvider() === sdk._meterProvider`, so the fallback resolves the SDK-owned provider.)

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  // Env span processors the SDK would build when the user configured none (zero-code). Resolved from
+  // the same sdk-node copy the SDK uses, so exporting keeps working after we add our processor.
+  function envSpanProcessors(): any[] {
+    try {
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const utils = require(require.resolve('@opentelemetry/sdk-node/build/src/utils', { paths: [sdkNodePath] }));
+      return utils.getSpanProcessorsFromEnv ? utils.getSpanProcessorsFromEnv() : [];
+    } catch {
+      return [];
```
liustve: **Swallowing all errors here into `[]` can silently drop *all* trace export.**

If this `catch` is ever hit, `envSpanProcessors()` returns `[]`, and the CONFIG normalizer then sets `spanProcessors = [SpanMetricsProcessor]` only (`src/internal/normalize-span-processors.ts:46-48`) — so NodeSDK exports **no spans** while 100% metrics keep flowing, masking the failure.

It is safe today (`getSpanProcessorsFromEnv` exists at 0.221 and returns a batch/OTLP processor by default), so this is a latent / defense-in-depth concern: this reaches into a non-public path (`@opentelemetry/sdk-node/build/src/utils`), so if that util is ever moved or renamed in a future `sdk-node`, span export dies silently.

Consider at minimum a `diag.warn` in the `catch` — a silent empty return is the one outcome that hides the regression — and/or preserving the SDK's own env fallback rather than forcing an empty list.

cloudwatch-plugin-otel/src/span-metrics.ts
```diff
@@ -0,0 +1,94 @@
...
+  } else if (config.traceExporter) {
+    // Convert the lone traceExporter to a processor (mirrors NodeSDK's default) and drop it, so the
+    // SDK does not ignore it once spanProcessors is set below.
+    processors = [new BatchSpanProcessor(config.traceExporter)];
+    delete config.traceExporter;
+  } else {
+    processors = [];
+  }
+  processors.push(new SpanMetricsProcessor());
+  config.spanProcessors = processors;
```
liustve: **Blocking: this `else` branch silently drops ALL span export in the env/default case — the programmatic twin of the register-path bug fixed for the singular/traceExporter cases.**

When a Mode-2 config carries no `spanProcessors`, no singular `spanProcessor`, and no `traceExporter`, this falls to `processors = []`, appends our processor, and sets `config.spanProcessors = [SpanMetricsProcessor]`. NodeSDK's precedence (`spanProcessors ?? [spanProcessor] ?? [batch(traceExporter)] ?? env`) then treats span processors as **explicitly configured** and skips its own env/default wiring — so the user's OTLP export is dropped and only the 100% RED metrics survive, masking the loss. This is the same failure class that was treated as must-fix on the register/CONFIG path (`envSpanProcessors()` + the `undefined`-abort contract in `normalize-span-processors.ts`).

Repro (real, **unpatched** sdk-node 0.221):

```
new NodeSDK({})                  -> [ BatchSpanProcessor<OTLPTraceExporter> ]   // export works
new NodeSDK(withSpanMetrics({})) -> [ SpanMetricsProcessor ]                    // has export path: false
```

Notes on reachability:
- **Config-driven, not env-driven.** `withSpanMetrics` never consults env for processors, so this drops export whether `OTEL_TRACES_EXPORTER` is unset (defaults to `otlp`) *or* set explicitly to `otlp`. It fires on any Mode-2 config lacking an explicit `traceExporter`/`spanProcessors`.
- **Legal, common pattern.** `new NodeSDK({ instrumentations })` relying on env/default OTLP is first-class OTel-JS usage; migrating it to `new NodeSDK(withSpanMetrics({ instrumentations }))` loses all traces silently. Every README example dodges this only because it happens to pass an explicit `traceExporter`; nothing warns and nothing documents that env/default export is unsupported in Mode 2.
- This is silent at runtime and contradicts this module's own stated principle of warning loudly rather than silently mis-wiring (`register.ts:11-14`).

**Suggested fix.** Unlike the register path, `withSpanMetrics` receives only the config object — it has no handle to the caller's `NodeSDK`/sdk-node copy, so it can't reliably resolve `getSpanProcessorsFromEnv` from the *same* sdk-node the caller constructs (a bare `require.resolve` would bind the plugin's own possibly-different-version copy). Given that, the safest fix here is **warn-and-abort** rather than mirroring the register path:

- In this `else` branch, emit a loud `diag.warn` — "no `traceExporter`/`spanProcessors` configured; appending the span-metrics processor would make NodeSDK treat processors as explicit and DROP env/default export; configure an explicit exporter" — and **do not set `config.spanProcessors`** (leave the config untouched so NodeSDK wires env/default itself). Span metrics stay off in that specific case (which the warning states), but the user keeps all their traces with zero coupling to the caller's sdk-node internals.
- A unit test should assert the resulting config **still carries an export processor** (not merely that our processor is present) — 100% line coverage of this pure transform doesn't catch this, because no existing test asserts the resulting NodeSDK still exports.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 02ef1c3 (hardened in b5703f1). Load-time failures now write to stderr unconditionally, in addition to diag.warn — at --require time no DiagLogger is installed yet, so stderr is the only guaranteed-visible channel. The app's logger is never replaced.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 3d99643 (extended in b5703f1). A resolution failure no longer forces an empty list: envSpanProcessors() returns undefined, the normalizer aborts without touching the config, and NodeSDK does its own env wiring — only the extension goes inert (warned, plus the post-start self-verify). An empty env list (OTEL_TRACES_EXPORTER=none) now also aborts, since upstream registers no tracer provider in that state.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 02ef1c3 and completed in b5703f1. patch() is wrapped in a top-level guard, and all imports (including @opentelemetry/api itself) are now required lazily inside patch(), so loading register.js executes nothing that can throw MODULE_NOT_FOUND. Verified with a standalone repro: node --require register.js in a directory with no OpenTelemetry packages exits 0 with a stderr notice.

cloudwatch-plugin-otel/src/internal/open-telemetry-holder.ts
srprash: Partially addressed in 5c83a1f: when nothing is explicitly bound, the holder now re-reads the global MeterProvider on every access (rejecting the noop), so a fresh provider registered globally after a restart is picked up. An explicitly bound provider remains first-wins; full shutdown-aware rebinding for that path is not implemented — treating in-process SDK restart with explicit bind() as unsupported for now.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 02ef1c3 and completed in b5703f1. patch() is guarded exactly as suggested, and — going one step further — every import (including @opentelemetry/api, where diag itself lives) is now required lazily inside patch(), so the guard also covers the module's own imports. The catch handler no longer assumes diag resolves; stderr is the unconditional channel. Tests cover missing api / sdk-trace-base / sdk-node via the resolution-hook repro, and the release workflow now smoke-tests the packed tarball's register entrypoint in a bare project.

cloudwatch-plugin-otel/src/span-metrics.ts
srprash: Fixed in dab3307 (refined in 6e36e61). withSpanMetrics now honors the singular spanProcessor with NodeSDK's own precedence — converted into spanProcessors, deleted so it cannot dangle, with a deprecation warning — matching the register path. When both plural and singular are set, the losing singular is deleted and the drop is warned about explicitly. Both paths now share a single resolution implementation (normalize-span-processors), so they cannot disagree on this input again.

cloudwatch-plugin-otel/contract-tests/apps/programmatic-app.ts
srprash: Fixed in bce4173. Both programmatic contract apps now exercise the no-bind flow (metricReaders in the NodeSDK config; the extension resolves the globally registered MeterProvider), and each wiring mode's contract test also asserts the metric resource carries the app's exact service.name — pinning the invariant the global-provider fallback relies on. The mischaracterizing comment was corrected; explicit bind() coverage lives in the manual mode, where the app genuinely owns a non-global provider. The README no longer labels any mode as recommended.

cloudwatch-plugin-otel/src/register.ts
srprash: Fixed in 3d99643. The catch no longer returns an empty list: envSpanProcessors() warns and returns undefined, and the normalizer treats that as an abort — config untouched, NodeSDK wires export from env upstream-style, only the extension goes inert. b5703f1 extends the same contract to a legitimately empty env list (OTEL_TRACES_EXPORTER=none), since upstream registers no tracer provider in that state and adding only our processor would have changed trace-context propagation.

.github/workflows/release-cloudwatch-plugin-otel.yml
srprash: This follows the same publish flow as the X-Ray UDP exporter in this repo (release-udp-exporter.yml: id-token: write, NPM_CONFIG_PROVENANCE, bare npm publish), which was proven on a brand-new package — @aws/aws-distro-opentelemetry-exporter-xray-udp did not exist on npm before its first release through that workflow. The bootstrap is the one-time org-side setup on npmjs.com (the @aws org account pre-authorizes the package/trusted publisher) before the first run; after that OIDC publishing works as-is. No workflow change needed.

cloudwatch-plugin-otel/src/span-metrics-attributes-builder.ts
srprash: The spec has been updated to resolve this contradiction: the attribute table is authoritative, and rpc.grpc.status_code is not part of the v1 allowlist. The code already follows the table, so no change is needed.

cloudwatch-plugin-otel/src/span-metrics-attributes-builder.ts
srprash: For v1 this stays as written: aws.otel.extension.lib.version is emitted on both spans and generated metrics. The value is the extension's own package version (build-time generated for JS). The possible move to a resource attribute remains a future spec revision; nothing in v1 depends on its placement (dedup keys on the schema attribute, not lib.version).

cloudwatch-plugin-otel/src/internal/normalize-span-processors.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    if (envProcessors.length === 0) {
+      // Same abort as R2: an empty env list (OTEL_TRACES_EXPORTER=none) means upstream registers no
+      // tracer provider; adding only our processor would create one and change propagation.
+      deps.warn(TRACE_EXPORT_DISABLED_MESSAGE);
+      return undefined;
+    }
+    processors = [...envProcessors];
+  }
+  processors.push(deps.makeSpanMetricsProcessor());
```
liustve, line 107: **Blocking:** This unconditionally appends a new `SpanMetricsProcessor`, even when the config already contains one. If `/register` is preloaded and the application also calls `withSpanMetrics()`, the patched `start()` normalizes the already-transformed config and appends a second processor. I reproduced one span producing `traces.span.metrics.calls = 2`, with `[UserProcessor, SpanMetricsProcessor, SpanMetricsProcessor]` active. This is realistic when a platform injects `NODE_OPTIONS` without the application owner knowing.

Please make registration idempotent. Use a shared `Symbol.for(...)` brand rather than `instanceof`, so the guard also works across multiple installed copies of this package. Please cover both `/register` + `withSpanMetrics()` and `withSpanMetrics(withSpanMetrics(config))`.

cloudwatch-plugin-otel/src/register.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+
+  // auto-instrumentations-node may resolve its OWN nested copy of sdk-node, which is a different
+  // module instance than a bare require('@opentelemetry/sdk-node') from here. Patch the SAME copy
+  // upstream will construct by resolving sdk-node relative to auto-instrumentations-node.
+  let sdkNodePath: string;
+  let pkgPath: string;
+  try {
+    const autoInstrEntry = require.resolve('@opentelemetry/auto-instrumentations-node');
```
liustve, line 68: **Blocking:** Resolving `sdk-node` relative to `auto-instrumentations-node` only guarantees that the upstream zero-code bootstrap uses the patched copy. If this preload is active but the application imports and constructs a different installed copy of `@opentelemetry/sdk-node`, `SpanMetricsNodeSDK` is never instantiated.

I reproduced this with the application resolving sdk-node 0.211 and the patch targeting 0.219: traces exported normally, the active list contained only `BatchSpanProcessor`, RED metrics stayed at zero, and no warning was emitted. The hook still logged `NodeSDK patched`, so this is a silent no-op reported as success.

Please add a late fail-safe that detects when the patched subclass was never constructed and warns that span metrics are disabled, for example a process-level flag set by the subclass constructor and checked at `beforeExit` (or an equivalent lifecycle point). A split-dependency-tree regression test should cover this case.

cloudwatch-plugin-otel/src/internal/normalize-span-processors.ts
liustve: This would be user setup issue, this is likely not common in practice

liustve (approved):

cloudwatch-plugin-otel/src/internal/normalize-span-processors.ts
srprash: This stems from a mixed-mode setup — the register preload and withSpanMetrics active in the same process — which is not an intended configuration: there is no scenario where both wiring paths should be enabled at once. Upstream OpenTelemetry does not guard against this either (register + a programmatic NodeSDK already produces duplicate processors/exporters and a global-provider conflict upstream; the OTel getting-started docs address it with a doc warning to avoid conflicting --require flags, not a runtime guard). Matching that posture, I'd like to drop this as a blocker and treat mixed-mode double-wiring as an unsupported configuration for now — we can revisit an idempotency brand in the future if real-world adoption shows this collision pattern is common.

cloudwatch-plugin-otel/src/register.ts
srprash: Same root scenario as the idempotency finding: the failure requires the register preload active while the application separately constructs NodeSDK from its own sdk-node copy — mixed-mode wiring. In pure zero-code use (the documented mode), upstream's register and the patch resolve sdk-node through the same path, so they always agree on the copy; the programmatic mode does not involve patching at all. Since mixed-mode instrumentation is not an intended use and upstream does not protect against it either, I'd like to drop this as a blocker and revisit later — e.g. a late fail-safe as you describe — if this setup turns out to be common in practice.

wangzlei (approved):

srprash:
Merging by bypassing the PR build success since the failures are due to CVE scan unrelated to this change.

PR #528, opened 2026-08-20
Carry platform fields onto the injected X-Ray trace header

### LINKED ISSUE #527: X-Ray context injection drops `Lineage`, disabling Lambda recursive loop detection
## Summary

`_adotInjectXrayContextMiddleware` replaces `X-Amzn-Trace-Id` wholesale on every AWS SDK
request. That displaces the AWS SDK's own `recursionDetectionMiddleware`, so the `Lineage`
field never leaves the function — and `Lineage` is how Lambda counts the hops of one
triggering event.

The effect is that Lambda's recursive loop detection stops working for chains through S3,
SQS and SNS. It fails silently: everything looks normal until a runaway loop is not
stopped.

## Where

`aws-distro-opentelemetry-node-autoinstrumentation/src/patches/smithy-send-patch.ts`:

```ts
this.middlewareStack?.add(
  (next: any) => async (middlewareArgs: any) => {
    propagation.inject(
      otelContext.active(),
      middlewareArgs.request.headers,
      defaultTextMapSetter,
    );
    const xrayId = middlewareArgs.request.headers[XRAY_TRACE_ID_HEADER];
    if (xrayId) {
      middlewareArgs.request.headers[XRAY_TRACE_ID_HEADER_CAPITALIZED] = xrayId;
      delete middlewareArgs.request.headers[XRAY_TRACE_ID_HEADER];
    }
    return await next(middlewareArgs);
  },
  { step: 'build', name: '_adotInjectXrayContextMiddleware', override: true },
);
```

Nothing here reads the header already on the request, or `_X_AMZN_TRACE_ID`. `inject`
builds a fresh `Root=..;Parent=..;Sampled=..` and sets it, so any other field that was
there is gone.

## Why it displaces the SDK's middleware

`recursionDetectionMiddleware` (`@aws-sdk/core`, `submodules/client`) is registered at
`step: "build"`, `priority: "low"`, and acts only when the header is absent:

```js
const traceIdHeader =
  Object.keys(request.headers ?? {}).find(
    (h) => h.toLowerCase() === X_AMZN_TRACE_ID.toLowerCase(),
  ) ?? X_AMZN_TRACE_ID;
if (!request.headers.hasOwnProperty(traceIdHeader)) {
  const traceId = traceIdFromInvokeStore ?? traceIdFromEnv;
  if (nonEmptyString(functionName) && nonEmptyString(traceId)) {
    request.headers[X_AMZN_TRACE_ID] = traceId; // this is what carries Lineage
  }
}
```

The middleware above specifies no `priority`, so it defaults to `normal`. Smithy sorts
within a step by descending priority weight (`@smithy/core`, `submodules/client`):

```js
const priorityWeights = { high: 3, normal: 2, low: 1 };
const sort = (entries) =>
  entries.sort(
    (a, b) =>
      stepWeights[b.step] - stepWeights[a.step] ||
      priorityWeights[b.priority || 'normal'] -
        priorityWeights[a.priority || 'normal'],
  );
```

So it runs first, sets the capitalised header, and `recursionDetectionMiddleware` then
finds it via its case-insensitive lookup and skips.

Note the ordering is not actually load-bearing for this bug. Were it reversed, the
`propagation.inject` above would overwrite the header the SDK had just written, discarding
`Lineage` just the same.

## Why `Lineage` matters

Per the Lambda documentation:

> To detect recursive loops, Lambda uses AWS X-Ray tracing headers... The updated metadata
> includes a count of the number of times that the event has invoked the function.
>
> If your function is invoked approximately 16 times in the same chain of requests, then
> Lambda automatically stops the next function invocation in that request chain.

Supported services are "your functions, Amazon SQS, Amazon S3, and Amazon SNS" — the same
integrations this middleware is most useful for. Feature is on by default for all accounts
and needs no X-Ray tracing enabled.

## Reproduction

Against `main`, driving the middleware this patch registers and asserting on the header it
leaves on the request. The test needs neither the AWS SDK nor a network stub — it takes the
middleware straight off the stack the patch adds to:

```
smithy send patch: X-Ray context injection
  ✓ injects the active trace under the capitalized header name
  1) carries Lineage from the environment onto the injected header
  2) prefers a header already on the request over the environment
  3) carries Self and unrecognised fields, and replaces the trace fields
  ✓ leaves the header as injected when there is nothing to carry

  2 passing, 3 failing
```

Each failure is the same shape:

```
Expected: "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1;Lineage=2:a87bd80c:1"
Received: "Root=1-5759e988-bd862e3fe1be46a994272793;Parent=53995c3f42cd8ad8;Sampled=1"
```

With `_X_AMZN_TRACE_ID` set to a header carrying `Lineage`, as the Lambda runtime sets it,
the request goes out without it. `Self` and unrecognised fields go the same way.

The test is included in the PR below, and passes 5 of 5 with the change applied.

Independently, the same defect was measured end to end on our own equivalent
implementation, across a `PutObject` → S3 notification → Lambda hop:

```
header replaced without carrying it : Lineage=1:aa20224c:0   (chain restarted every hop)
header replaced carrying it forward : Lineage=2:aa20224c:0   (chain counted)
```

## Suggested fix

Carry every field that is not `Root`/`Parent`/`Sampled` from the header being replaced onto
the new one. `Lineage` is the one that matters operationally; `Self`, which an Application
Load Balancer stamps, has the same character, and ALB documents the general convention:

> An application can add arbitrary fields for its own purposes. The load balancer preserves
> these fields but does not use them.

A PR follows. It is contained to the one middleware, adds a test that drives it directly, and
does not change the trace context that gets injected.

artkhod:
Fixes #527.

## Problem

`_adotInjectXrayContextMiddleware` replaces `X-Amzn-Trace-Id` wholesale. `propagation.inject`
builds a fresh `Root=..;Parent=..;Sampled=..` and sets it, so any other field that was on the
header is gone.

The field that matters is `Lineage`. Lambda counts the hops of one triggering event in it and
stops invoking a function after roughly sixteen in a chain, through Lambda, S3, SQS and SNS.
Because this middleware defaults to `normal` priority and the AWS SDK's own
`recursionDetectionMiddleware` sits on the same `build` step at `low`, this one runs first;
that middleware then finds the header already present and skips, so the count never leaves
the function. Recursive loop detection stops working, and it fails silently — nothing looks
wrong until a runaway loop is not stopped.

## Change

`src/patches/smithy-send-patch.ts` — the injection middleware now reads the header it is
about to replace before replacing it, and appends whatever it held that is not
`Root`/`Parent`/`Sampled`. Two small helpers are added for that; the trace context injected
is unchanged.

The header already on the request is preferred, falling back to `_X_AMZN_TRACE_ID`, so a
caller that set one explicitly is respected.

Unrecognised fields are carried too, not only `Lineage` and `Self`. That is the convention on
this header — an Application Load Balancer documents that "an application can add arbitrary
fields for its own purposes. The load balancer preserves these fields but does not use them"
— and it means a field AWS adds later needs no change here.

## Why read the environment variable rather than order after the SDK's middleware

Ordering after `recursionDetectionMiddleware` would mean dropping this middleware to `low`
priority, which moves it relative to everything else on the `build` step. The environment
variable is where that middleware reads from anyway, so this produces the same value without
coupling the two by ordering.

## Tests

Four cases are added to the existing `overridden _getV3SmithyClientSendPatch updates
MiddlewareStack` block in `test/patches/instrumentation-patch.test.ts`, following the sinon
stub already used there for `AWSXRayPropagator.prototype.inject` — so they exercise the
middleware without depending on the globally configured propagator.

- carries `Lineage` from `_X_AMZN_TRACE_ID` onto the injected header
- prefers a header already on the request over the environment variable
- carries `Self` and unrecognised fields, and does not carry the replaced trace fields
- leaves the injected header alone when there is nothing to carry

That file goes from 46 passing / 5 failing before the change to 49 passing / 2 failing after.

## The 2 remaining failures are pre-existing

`prevents recursion when credentials provider makes STS calls` and `injects trace context
header into request via propagator` fail identically with and without this change on a clean
checkout here. Both stop at `expect(...).toBe(1)` receiving `0` — no span is captured, so
neither reaches a header assertion. That looks environmental rather than related to this
area, but flagging it since one is named after the middleware this PR touches.

## Compatibility

No configuration change, and no change to the trace context injected. The only difference is
that fields the middleware previously discarded now survive, which restores the behaviour a
caller would have had without ADOT installed.


wangzlei:
Nice catch on this one — the write-up made the root cause really easy to follow, and the fix looks correct. One heads-up: longer term we're trying to move X-Ray propagation toward W3C headers and stop stuffing extra fields into X-Amzn-Trace-Id. The Java side has a good reference for where we're headed https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1671

Mind taking a peek at that PR and seeing if we can line this up with the same model -- fixing from propagator instead of Middleware

artkhod:
> Nice catch on this one — the write-up made the root cause really easy to follow, and the fix looks correct. One heads-up: longer term we're trying to move X-Ray propagation toward W3C headers and stop stuffing extra fields into X-Amzn-Trace-Id. The Java side has a good reference for where we're headed [open-telemetry/opentelemetry-java-contrib#1671](https://github.com/open-telemetry/opentelemetry-java-contrib/pull/1671)
> 
> Mind taking a peek at that PR and seeing if we can line this up with the same model -- fixing from propagator instead of Middleware

I think there's still argument to have X-ray propagation around, and that is lambda -> S3, S3 event notification -> lambda. S3 event notification does not include any metadata, but with X-ray propagation the link survives (unfortunately, there seems to be no good solution to propagate baggage here). I have also raised an issue & PR in aws-sdk instrumentation (linked in the issue in this repo as well) to use X-ray propagation for that exact reason.

PR #531, opened 2026-08-21
fix(agent-observability): suppress invalid instrumentation startup logs

liustve:
## Summary
- suppress false missing-instrumentation errors for ADOT GenAI instrumentations
- clarify Vercel AI span processor registration logs

## Logs
```text
Provided instrumentation name "@opentelemetry/instrumentation-aws_langchain" not found
Provided instrumentation name "@opentelemetry/instrumentation-aws_openai_agents" not found
Provided instrumentation name "@opentelemetry/instrumentation-aws_vercel_ai" not found
@aws/aws-distro-opentelemetry-instrumentation-vercel-ai Failed to register VercelAISpanProcessor
```

## Testing
- `npm test`
- `npm run compile --workspace @aws/aws-distro-opentelemetry-node-autoinstrumentation`


aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
```diff
@@ -43,9 +43,28 @@ import {
...
+    // https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e52a9053029304f271b7dbe1b07e7fb2b987e30/packages/auto-instrumentations-node/src/utils.ts#L226-L232
+    if (
+      ![LANGCHAIN_SHORT_NAME, OPENAI_AGENTS_SHORT_NAME, VERCEL_AI_SHORT_NAME]
+        .map(name => `Provided instrumentation name "@opentelemetry/instrumentation-${name}" not found`)
+        .includes(message)
+    ) {
+      consoleLogger.error(message, ...args);
+    }
+  },
+};
```
jj22ee: This is heavier than it needs to be, in two ways:

1. The four hand-copied methods are unnecessary — `DiagConsoleLogger` assigns `error`/`warn`/`info`/`debug`/`verbose` as **own instance properties** in its constructor ([consoleLogger.ts#L72-L74](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/diag/consoleLogger.ts#L72-L74)), so a spread covers them.
2. The `.map()` runs on **every** `diag.error` call for the lifetime of the process, rebuilding three strings to do a linear compare. Build the set once at module load.

```suggestion
// Upstream logs an error for every name in OTEL_NODE_{ENABLED,DISABLED}_INSTRUMENTATIONS that
// isn't in its own instrumentation map, which includes our GenAI instrumentations:
// https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e52a9053029304f271b7dbe1b07e7fb2b987e30/packages/auto-instrumentations-node/src/utils.ts#L226-L232
const UNKNOWN_INSTRUMENTATION_ERRORS = new Set(
  [LANGCHAIN_SHORT_NAME, OPENAI_AGENTS_SHORT_NAME, VERCEL_AI_SHORT_NAME].map(
    name => `Provided instrumentation name "@opentelemetry/instrumentation-${name}" not found`
  )
);

const consoleLogger = new DiagConsoleLogger();
const diagLogger: DiagLogger = {
  ...consoleLogger,
  error: (message: string, ...args: unknown[]): void => {
    if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
      consoleLogger.error(message, ...args);
    }
  },
};
```

Note for whoever touches this next: `class extends DiagConsoleLogger { override error() {...} }` does *not* work — the base constructor's own property shadows the subclass prototype method. So the wrapper-object shape is the right approach here, it's just over-typed as written.

Two alternatives I looked at and would **not** recommend:

- Scoping the filter to just the `getNodeAutoInstrumentations()` call on L266 (the only place this error is emitted) would bound it in time, but it costs more code, not less: re-calling `diag.setLogger` emits "Current logger will be overwritten..." warnings unless you also pass `suppressOverrideMessage: true`.
- Stripping the `aws_*` names from the env var before the upstream call is a trap: if a user sets `OTEL_NODE_ENABLED_INSTRUMENTATIONS` to *only* GenAI names, the filtered value is empty, and deleting the var flips upstream from "enable nothing" to "enable everything."

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-vercel-ai/instrumentation.ts
```diff
@@ -68,15 +68,16 @@ export class VercelAIInstrumentation extends InstrumentationBase<VercelAIInstrum
 
   override setTracerProvider(provider: any): void {
     if (!this._spanProcessorRegistered) {
+      this._diag.debug('Registering VercelAISpanProcessor');
       const delegate = provider.getDelegate?.() ?? provider;
       const processors = delegate._activeSpanProcessor?._spanProcessors;
 
       if (!Array.isArray(processors)) {
-        this._diag.warn('Failed to register VercelAISpanProcessor');
+        this._diag.debug('Failed to register VercelAISpanProcessor');
```
jj22ee: Downgrading this to `debug` makes a real failure silent. Without `VercelAISpanProcessor` registered, Vercel AI spans are never translated into GenAI semconv, so the customer gets spans carrying Vercel's own attributes and no signal explaining why.

I understand why it was downgraded — `setTracerProvider` is called speculatively and the early call can legitimately miss (hence the `!this._spanProcessorRegistered` guard), so `warn` here was a false positive. But the fix for a false-positive warning shouldn't also delete the true positive.

Suggestion: keep per-attempt misses at `debug` as done here, and `warn` once at a point where the failure is unambiguous — e.g. in `_enableTelemetryByDefaultWrapper`, if the `ai` module is actually being patched while `!this._spanProcessorRegistered`, that's a genuine problem worth surfacing.

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-vercel-ai/instrumentation.ts
```diff
@@ -68,15 +68,16 @@ export class VercelAIInstrumentation extends InstrumentationBase<VercelAIInstrum
...
       } else {
         processors.unshift(new VercelAISpanProcessor());
         this._spanProcessorRegistered = true;
-        this._diag.debug('Registered VercelAISpanProcessor');
+        this._diag.info('Registered VercelAISpanProcessor successfully');
```
jj22ee: This prints on **every** startup for **every** app, which works against the goal of the PR.

Default diag level is INFO when `OTEL_LOG_LEVEL` is unset, and [register.ts#L288-L294](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts#L288-L294) calls `setTracerProvider` on every instrumentation in the list unconditionally — regardless of whether `ai` is installed, and regardless of `isEnabled()`. So a PR whose purpose is suppressing startup noise adds an unconditional startup line for users who never touch Vercel AI.

I'd leave this at `debug`.

```suggestion
        this._diag.debug('Registered VercelAISpanProcessor successfully');
```

Separately (pre-existing, not this PR's job): because `setTracerProvider` doesn't check enabled state, the span processor is registered even when the user sets `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai`. Worth a follow-up.

aws-distro-opentelemetry-node-autoinstrumentation/test/register.test.ts
```diff
@@ -118,6 +118,30 @@ describe('Register', function () {
...
+      const configuredNames = `${customNames},${invalidName}`;
+
+      const proc = spawnWithAssertion(
+        {
+          OTEL_LOG_LEVEL: 'INFO',
+          OTEL_NODE_ENABLED_INSTRUMENTATIONS: configuredNames,
+          OTEL_NODE_DISABLED_INSTRUMENTATIONS: configuredNames,
+        },
+        `assert.strictEqual(process.env.OTEL_NODE_ENABLED_INSTRUMENTATIONS, '${configuredNames}');
+         assert.strictEqual(process.env.OTEL_NODE_DISABLED_INSTRUMENTATIONS, '${configuredNames}');`
```
jj22ee: These two assertions have nothing to do with "suppresses only custom instrumentation registry errors" — the actual suppression assertions are the stderr checks below.

They are asserting a real invariant (register.ts must not permanently mutate these vars, since `isInstrumentationDisabled()` reads them later at `enable()` time), so I'd keep them — but under their own test name rather than riding along here.

One coverage gap while you're in here: nothing asserts the filter is scoped to `error` specifically, i.e. that an unrelated `diag.error` still reaches stderr. The `not-registered` case covers that only partially.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ If your change does not need a CHANGELOG entry, add the "skip changelog" label t
 
 ## Unreleased
 
+- fix(agent-observability): suppress invalid instrumentation startup logs
```
jj22ee: Missing the PR link that every neighboring entry carries.

```suggestion
- fix(agent-observability): suppress invalid instrumentation startup logs
  ([#531](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/531))
```

aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
jj22ee: This is heavier than it needs to be, in two ways:

1. The four hand-copied methods are unnecessary — `DiagConsoleLogger` assigns `error`/`warn`/`info`/`debug`/`verbose` as **own instance properties** in its constructor ([consoleLogger.ts#L72-L74](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/diag/consoleLogger.ts#L72-L74)), so a spread covers them.
2. The `.map()` runs on **every** `diag.error` call for the lifetime of the process, rebuilding three strings to do a linear compare. Build the set once at module load.

```suggestion
// Upstream logs an error for every name in OTEL_NODE_{ENABLED,DISABLED}_INSTRUMENTATIONS that
// isn't in its own instrumentation map, which includes our GenAI instrumentations:
// https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e52a9053029304f271b7dbe1b07e7fb2b987e30/packages/auto-instrumentations-node/src/utils.ts#L226-L232
const UNKNOWN_INSTRUMENTATION_ERRORS = new Set(
  [LANGCHAIN_SHORT_NAME, OPENAI_AGENTS_SHORT_NAME, VERCEL_AI_SHORT_NAME].map(
    name => `Provided instrumentation name "@opentelemetry/instrumentation-${name}" not found`
  )
);

const consoleLogger = new DiagConsoleLogger();
const diagLogger: DiagLogger = {
  ...consoleLogger,
  error: (message: string, ...args: unknown[]): void => {
    if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
      consoleLogger.error(message, ...args);
    }
  },
};
```

Note for whoever touches this next: `class extends DiagConsoleLogger { override error() {...} }` does *not* work — the base constructor's own property shadows the subclass prototype method. So the wrapper-object shape is the right approach here, it's just over-typed as written.

Two alternatives I looked at and would **not** recommend:

- Scoping the filter to just the `getNodeAutoInstrumentations()` call on L266 (the only place this error is emitted) would bound it in time, but it costs more code, not less: re-calling `diag.setLogger` emits "Current logger will be overwritten..." warnings unless you also pass `suppressOverrideMessage: true`.
- Stripping the `aws_*` names from the env var before the upstream call is a trap: if a user sets `OTEL_NODE_ENABLED_INSTRUMENTATIONS` to *only* GenAI names, the filtered value is empty, and deleting the var flips upstream from "enable nothing" to "enable everything."

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-vercel-ai/instrumentation.ts
```diff
@@ -68,15 +68,16 @@ export class VercelAIInstrumentation extends InstrumentationBase<VercelAIInstrum
...
       }
```
jj22ee: These two severity changes invert the signal: the failure is now invisible and the success is now the loudest line in startup. I don't think `debug` is the right fix for either.

**The false positive is real — I confirmed it, and it fires on every single startup.** But it's a *placement* problem, not a volume problem:

1. `sdk.start()` calls `registerInstrumentations()` as its very first statement ([sdk-node@0.219.0 sdk.js#L186](https://unpkg.com/@opentelemetry/sdk-node@0.219.0/build/src/sdk.js)).
2. No `tracerProvider` is passed, so it falls back to `trace.getTracerProvider()` — the still-unbound `ProxyTracerProvider` — and calls `setTracerProvider` on every instrumentation ([autoLoader.js#L18](https://unpkg.com/@opentelemetry/instrumentation@0.219.0/build/src/autoLoader.js)).
3. `getDelegate()` returns the noop provider, which has no `_activeSpanProcessor`, so `processors` is `undefined` → the warning fires.
4. The real `NodeTracerProvider` isn't constructed until sdk.js:231 and only registered globally at :237. ADOT then re-calls `setTracerProvider` after `start()` ([register.ts#L288-L294](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts#L288-L294)), which succeeds — that's exactly what the `_spanProcessorRegistered` guard exists for.

So failure is *expected* on the first call and the code already knows it. Lowering the severity silences the expected case and the unexpected case together, and the unexpected case matters: `_activeSpanProcessor._spanProcessors` are private SDK internals. If upstream renames or restructures them, registration silently stops working, every Vercel AI span quietly loses its GenAI semconv translation, and there is no signal at any log level. That's a bad failure mode to buy for a startup line.

The `info` on the success path has the same problem in reverse: default diag level is INFO when `OTEL_LOG_LEVEL` is unset, and register.ts calls `setTracerProvider` on every instrumentation unconditionally — regardless of whether `ai` is installed or whether `isEnabled()` is true. So a PR whose goal is suppressing startup noise adds an unconditional startup line for every app, including the ones that never touch Vercel AI.

Drop the log from the branch where failure is expected, and keep success at `debug`:

```suggestion
      if (Array.isArray(processors)) {
        processors.unshift(new VercelAISpanProcessor());
        this._spanProcessorRegistered = true;
        this._diag.debug('Registered VercelAISpanProcessor');
      } else {
        // Expected on the pre-start call from registerInstrumentations(), where the global
        // tracer provider is still an unbound proxy. The post-start call from register.ts
        // is the one that has to succeed.
        this._diag.debug('Tracer provider has no active span processors yet');
      }
```

Then warn once at the point where the failure is unambiguous — in `_enableTelemetryByDefaultWrapper`, where we know the `ai` module is really being patched:

```ts
if (!this._spanProcessorRegistered) {
  this._diag.warn(
    'VercelAISpanProcessor is not registered; Vercel AI spans will not be translated to OTel GenAI semantic conventions'
  );
}
```

That removes the noise you were after *and* keeps the signal. (For completeness: the other branch that reaches the `else` is `spanProcessors.length === 0` in sdk.js:229-238, where no global provider is ever set — but with zero span processors there's no tracing at all, so it's moot.)

Separately (pre-existing, not this PR's job): because `setTracerProvider` doesn't check enabled state, the span processor is registered even when the user sets `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai`. Worth a follow-up.

aws-distro-opentelemetry-node-autoinstrumentation/test/register.test.ts
jj22ee: These two assertions have nothing to do with "suppresses only custom instrumentation registry errors" — the actual suppression assertions are the stderr checks below.

They are asserting a real invariant (register.ts must not permanently mutate these vars, since `isInstrumentationDisabled()` reads them later at `enable()` time), so I'd keep them — but under their own test name rather than riding along here.

One coverage gap while you're in here: nothing asserts the filter is scoped to `error` specifically, i.e. that an unrelated `diag.error` still reaches stderr. The `not-registered` case covers that only partially.

CHANGELOG.md
jj22ee: Missing the PR link that every neighboring entry carries.

```suggestion
- fix(agent-observability): suppress invalid instrumentation startup logs
  ([#531](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/531))
```

aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
jj22ee: This is heavier than it needs to be, in two ways:

1. The four hand-copied methods are unnecessary — `DiagConsoleLogger` assigns `error`/`warn`/`info`/`debug`/`verbose` as **own instance properties** in its constructor ([consoleLogger.ts#L72-L74](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/diag/consoleLogger.ts#L72-L74)), so a spread covers them.
2. The `.map()` runs on **every** `diag.error` call for the lifetime of the process, rebuilding three strings to do a linear compare. Build the set once at module load.

```suggestion
// Upstream logs an error for every name in OTEL_NODE_{ENABLED,DISABLED}_INSTRUMENTATIONS that
// isn't in its own instrumentation map, which includes our GenAI instrumentations:
// https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e52a9053029304f271b7dbe1b07e7fb2b987e30/packages/auto-instrumentations-node/src/utils.ts#L226-L232
const UNKNOWN_INSTRUMENTATION_ERRORS = new Set(
  [LANGCHAIN_SHORT_NAME, OPENAI_AGENTS_SHORT_NAME, VERCEL_AI_SHORT_NAME].map(
    name => `Provided instrumentation name "@opentelemetry/instrumentation-${name}" not found`
  )
);

const consoleLogger = new DiagConsoleLogger();
const diagLogger: DiagLogger = {
  ...consoleLogger,
  error: (message: string, ...args: unknown[]): void => {
    if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
      consoleLogger.error(message, ...args);
    }
  },
};
```

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-vercel-ai/instrumentation.ts
jj22ee: These two severity changes invert the signal: the failure is now invisible and the success is now the loudest line in startup. I don't think `debug` is the right fix for either.

**The false positive is real — I confirmed it, and it fires on every single startup.** But it's a *placement* problem, not a volume problem:

1. `sdk.start()` calls `registerInstrumentations()` as its very first statement ([sdk-node@0.219.0 sdk.js#L186](https://unpkg.com/@opentelemetry/sdk-node@0.219.0/build/src/sdk.js)).
2. No `tracerProvider` is passed, so it falls back to `trace.getTracerProvider()` — the still-unbound `ProxyTracerProvider` — and calls `setTracerProvider` on every instrumentation ([autoLoader.js#L18](https://unpkg.com/@opentelemetry/instrumentation@0.219.0/build/src/autoLoader.js)).
3. `getDelegate()` returns the noop provider, which has no `_activeSpanProcessor`, so `processors` is `undefined` → the warning fires.
4. The real `NodeTracerProvider` isn't constructed until sdk.js:231 and only registered globally at :237. ADOT then re-calls `setTracerProvider` after `start()` ([register.ts#L288-L294](https://github.com/aws-observability/aws-otel-js-instrumentation/blob/main/aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts#L288-L294)), which succeeds — that's exactly what the `_spanProcessorRegistered` guard exists for.

So failure is *expected* on the first call and the code already knows it. Lowering the severity silences the expected and unexpected cases together, and the unexpected case matters: `_activeSpanProcessor._spanProcessors` are private SDK internals. If upstream renames or restructures them, registration silently stops working, every Vercel AI span quietly loses its GenAI semconv translation, and there is no signal at any log level. That's a bad failure mode to buy for a startup line.

The `info` on the success path has the same problem in reverse: default diag level is INFO when `OTEL_LOG_LEVEL` is unset, and register.ts calls `setTracerProvider` on every instrumentation unconditionally — regardless of whether `ai` is installed or whether `isEnabled()` is true. So a PR whose goal is suppressing startup noise adds an unconditional startup line for every app, including the ones that never touch Vercel AI.

Drop the log from the branch where failure is expected, and keep success at `debug`:

```suggestion
      if (Array.isArray(processors)) {
        processors.unshift(new VercelAISpanProcessor());
        this._spanProcessorRegistered = true;
        this._diag.debug('Registered VercelAISpanProcessor');
      } else {
        // Expected on the pre-start call from registerInstrumentations(), where the global
        // tracer provider is still an unbound proxy. The post-start call from register.ts
        // is the one that has to succeed.
        this._diag.debug('Tracer provider has no active span processors yet');
      }
```

Then warn once at the point where the failure is unambiguous — in `_enableTelemetryByDefaultWrapper`, where we know the `ai` module is really being patched:

```ts
if (!this._spanProcessorRegistered) {
  this._diag.warn(
    'VercelAISpanProcessor is not registered; Vercel AI spans will not be translated to OTel GenAI semantic conventions'
  );
}
```

That removes the noise you were after *and* keeps the signal.

Separately (pre-existing, not this PR's job): because `setTracerProvider` doesn't check enabled state, the span processor is registered even when the user sets `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai`. Worth a follow-up.

aws-distro-opentelemetry-node-autoinstrumentation/test/register.test.ts
jj22ee: Nit: this name points at the diag filter, but these two assertions are about something else — that register.ts doesn't *permanently* mutate the env vars, which matters because `isInstrumentationDisabled()` reads them later at `enable()` time. If they ever fail, the reported test name sends the next reader to the wrong place.

Renaming the test to cover both is enough to fix that. Splitting the env-mutation invariant into its own test would be my preference — `spawnWithAssertion` accepts `''` as the assertion, and `assert.equal(proc.status, 0)` still verifies register.js loaded — but that costs another `spawnSync` in a file that already spawns a lot, so I won't push on it.

CHANGELOG.md
jj22ee: Missing the PR link that every neighboring entry carries.

```suggestion
- fix(agent-observability): suppress invalid instrumentation startup logs
  ([#531](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/531))
```

aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
jj22ee: Two optimizations:

1. The copied methods are unnecessary — `DiagConsoleLogger` assigns all five as own instance properties in its constructor ([consoleLogger.ts#L72-L74](https://github.com/open-telemetry/opentelemetry-js/blob/main/api/src/diag/consoleLogger.ts#L72-L74)), so a spread covers them.
2. The `.map()` rebuilds three strings on every `diag.error` call for the life of the process. Build the set once.

```suggestion
// Upstream logs an error for every name in OTEL_NODE_{ENABLED,DISABLED}_INSTRUMENTATIONS that
// isn't in its own instrumentation map, which includes our GenAI instrumentations:
// https://github.com/open-telemetry/opentelemetry-js-contrib/blob/4e52a9053029304f271b7dbe1b07e7fb2b987e30/packages/auto-instrumentations-node/src/utils.ts#L226-L232
const UNKNOWN_INSTRUMENTATION_ERRORS = new Set(
  [LANGCHAIN_SHORT_NAME, OPENAI_AGENTS_SHORT_NAME, VERCEL_AI_SHORT_NAME].map(
    name => `Provided instrumentation name "@opentelemetry/instrumentation-${name}" not found`
  )
);

const consoleLogger = new DiagConsoleLogger();
const diagLogger: DiagLogger = {
  ...consoleLogger,
  error: (message: string, ...args: unknown[]): void => {
    if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
      consoleLogger.error(message, ...args);
    }
  },
};
```

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-vercel-ai/instrumentation.ts
jj22ee, line 83: Could we put the success line back to `debug`? INFO is the default level when `OTEL_LOG_LEVEL` is unset, and register.ts calls `setTracerProvider` on every instrumentation unconditionally — regardless of whether `ai` is installed or whether the instrumentation is disabled — so this prints on every startup for every app, which works against the goal of the PR. It's also inconsistent with the siblings: langchain and openai-agents use `info` only for the third-party-conflict skip, never to announce successful setup.

The `warn` → `debug` on the failure path is fine by me — that branch only ever sees the expected pre-start call, since `sdk.start()` registers instrumentations against the still-unbound `ProxyTracerProvider` before creating the real provider, and the post-start call in register.ts is the one that succeeds.

Here's a suggested re-wording of the debug log on L76, since "Failed to register" reads like a real failure when it's just the expected first pass:

```ts
this._diag.debug('Skipping VercelAISpanProcessor registration: tracer provider has no active span processors yet');
```

You might want to add a test for this — nothing covers the log behavior today. The existing `Vercel AI auto-registers VercelAISpanProcessor` test (register.test.ts:277) only asserts that registration happens.

Unrelated to your change, but I noticed it while reading: `setTracerProvider` registers the span processor without checking `isEnabled()`, so `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai` doesn't actually stop `VercelAISpanProcessor` from being added and rewriting span attributes. Worth a separate issue?

CHANGELOG.md
jj22ee, line 16:
```suggestion
- fix(agent-observability): suppress invalid instrumentation startup logs
  ([#531](https://github.com/aws-observability/aws-otel-js-instrumentation/pull/531))
```

aws-distro-opentelemetry-node-autoinstrumentation/src/register.ts
```diff
@@ -43,9 +43,28 @@ import {
...
+
+const consoleLogger = new DiagConsoleLogger();
+const diagLogger: DiagLogger = {
+  ...consoleLogger,
+  error: (message: string, ...args: unknown[]): void => {
+    if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
+      consoleLogger.error(message, ...args);
+    }
+  },
+};
```
jj22ee: Non-blocking, and a refinement of my own earlier suggestion rather than a new problem.

`{...consoleLogger}` produces a plain object, so `diagLogger` is no longer a `DiagConsoleLogger` — `instanceof` is false and the prototype is `Object.prototype`. Nothing checks that, so it's fine today. But it works only because `DiagConsoleLogger` assigns all five methods as own instance properties in its constructor. If upstream ever moves them to the prototype, the spread yields an object with no `verbose`/`debug`/`info`/`warn`, and `createLogLevelDiagLogger` quietly substitutes no-ops (`typeof theFunc === 'function' ? theFunc.bind(logger) : function () {}`), so those levels would just stop logging. TypeScript wouldn't catch it either, since the class declares the properties with definite assignment (`public error!: DiagLogFunction`).

Patching the instance drops that assumption — it only touches `error`, and the `.bind` keeps it independent of whether the underlying methods use `this`:

```suggestion
const diagLogger: DiagLogger = new DiagConsoleLogger();
const consoleError = diagLogger.error.bind(diagLogger);
diagLogger.error = (message: string, ...args: unknown[]): void => {
  if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
    consoleError(message, ...args);
  }
};
```

Both forms are correct against the current implementation, so happy to leave it as-is if you prefer not mutating the instance — the class JSDoc does describe it as immutable, even though nothing is frozen.

jj22ee: A refinement of my own earlier suggestion.

`{...consoleLogger}` produces a plain object, so `diagLogger` is no longer a `DiagConsoleLogger` — `instanceof` is false and the prototype is `Object.prototype`. Nothing checks that, so it's fine today.

Patching the instance is better, here it only touches `error`, and the `.bind` keeps it independent of whether the underlying methods use `this`:

```suggestion
const diagLogger: DiagLogger = new DiagConsoleLogger();
const consoleError = diagLogger.error.bind(diagLogger);
diagLogger.error = (message: string, ...args: unknown[]): void => {
  if (!UNKNOWN_INSTRUMENTATION_ERRORS.has(message)) {
    consoleError(message, ...args);
  }
};
```

jj22ee (approved):
LGTM.

One thing that's out of scope for this PR: `setTracerProvider` registers `VercelAISpanProcessor` without checking `isEnabled()`, so `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai` doesn't stop the processor from being added. It only bites when the app enables `experimental_telemetry` itself — then those spans still get rewritten despite the instrumentation being disabled. Could that be picked up in a follow-up PR?

liustve:
> LGTM.
> 
> One thing that's out of scope for this PR: `setTracerProvider` registers `VercelAISpanProcessor` without checking `isEnabled()`, so `OTEL_NODE_DISABLED_INSTRUMENTATIONS=aws_vercel_ai` doesn't stop the processor from being added. It only bites when the app enables `experimental_telemetry` itself — then those spans still get rewritten despite the instrumentation being disabled. Could that be picked up in a follow-up PR?

will do a followup PR

srprash (approved):

PR #540, opened 2026-08-28
add post release artifact validation

haneric00:
title

As part of adding trivy artifact verification to release builds.


.github/workflows/release-build.yml
```diff
@@ -162,16 +158,32 @@ jobs:
...
+            -t "${{ env.RELEASE_PRIVATE_REPOSITORY }}:v${{ env.VERSION }}" \
+            "$STAGING_IMAGE"
+
+      - name: Perform image scan
+        uses: ./.github/actions/image_scan
         with:
-          push: true
-          context: .
-          file: ./Dockerfile
-          platforms: linux/amd64,linux/arm64
```
v-iw, line 171: [P1]
Currently: This replaces the previous release-time linux/amd64,linux/arm64 build with promotion of the image produced by main build.

Gap: The existing staging build in 
action.yml 106
 is hardcoded to platforms: linux/amd64. The unchanged release logic later requires both amd64 and arm64 and exits if either is missing. Therefore, the staging image is not currently equivalent to the multi-architecture image produced by the removed release step.

Request: Could the composite action accept a platforms input and have main build pass linux/amd64,linux/arm64? PR/local builds that use load: true can continue passing only linux/amd64, while the pushed staging image should contain both release architectures.

.github/workflows/release-build.yml
```diff
@@ -162,16 +158,32 @@ jobs:
...
+      # Promote the exact multi-arch image the main build already built, scanned, and verified:
+      # copy its manifest (all arches + attestations) from staging to the release registries by
+      # reference -- no rebuild, no re-verify. The staging tag is reconstructed the same way
+      # main-build.yml builds it: <root package.json version>-<short sha> of this (same) commit.
+      - name: Promote staging image to release ECRs
+        run: |
+          set -euo pipefail
+          PKG_VERSION=$(jq -r '.version' package.json)
+          SHORT_SHA=$(git rev-parse --short HEAD)
+          STAGING_IMAGE="${{ env.STAGING_ECR_REGISTRY }}/${{ env.STAGING_ECR_REPOSITORY }}:${PKG_VERSION}-${SHORT_SHA}"
```
v-iw: [P1]
Currently: The npm tarball is downloaded from a selected main-build run, but the image reference is independently reconstructed from the release checkout’s package version and short SHA.

Gap: These may identify different commits, and a same-commit rerun may reuse the staging tag. Therefore, release is not guaranteed to promote the image validated by the run that produced the tarball.

Request: Could JavaScript follow the updated Python approach: record the pushed multi-arch image digest in main build, upload the digest-qualified reference as an artifact, and promote that reference from the same selected run as the .tgz?

Issue #548, opened 2026-09-08
LangChain instrumentation crashes with MODULE_NOT_FOUND when injected by the Kubernetes operator


jtodaroii:
## Summary

When the distro is injected by the OpenTelemetry Operator (as the `amazon-cloudwatch-observability` EKS add-on does), the LangChain instrumentation resolves `@langchain/core` from its own mount at `/otel-auto-instrumentation-nodejs` instead of the application's `node_modules`. On any app using `@langchain/core@1.x`, this throws `MODULE_NOT_FOUND` into the request path and breaks every LLM call.

The distro works when installed as an application dependency, because `@langchain/core` is then a resolvable sibling. It breaks in the operator's volume-mount layout, where it is not.

## Environment

- `@aws/aws-distro-opentelemetry-node-autoinstrumentation` **0.11.0** (image `adot-autoinstrumentation-node:v0.11.0`)
- EKS add-on `amazon-cloudwatch-observability` **v6.2.0-eksbuild.1**
- Node **24.16.0**, `@langchain/core` **1.2.0**
- Injected via `NODE_OPTIONS=--require /otel-auto-instrumentation-nodejs/autoinstrumentation.js`

## What happens

Every model invocation fails:

```
Error: Cannot find module '@langchain/core/callbacks/base'
Require stack:
- /otel-auto-instrumentation-nodejs/node_modules/@aws/aws-distro-opentelemetry-node-autoinstrumentation/build/src/instrumentation/instrumentation-langchain/callback-handler.js
- /otel-auto-instrumentation-nodejs/node_modules/@aws/aws-distro-opentelemetry-node-autoinstrumentation/build/src/instrumentation/instrumentation-langchain/instrumentation.js
- ...
- /otel-auto-instrumentation-nodejs/autoinstrumentation.js
- internal/preload
```

The error surfaces at the first `CallbackManager._configureSync`, not at startup, because `enable()` loads the handler lazily.

## Root cause

The distro is mounted at `/otel-auto-instrumentation-nodejs`, while the app's packages live at `/app/node_modules`. Node's resolution from the distro's own directory walks up to `/otel-auto-instrumentation-nodejs/node_modules` and then `/node_modules` and stops. Verified inside a running pod:

```
/app/node_modules/@langchain/core                  -> present, 1.2.0, callbacks/base.cjs present
/otel-auto-instrumentation-nodejs/node_modules/@langchain  -> does not exist
/node_modules                                      -> does not exist
[...]

PR #549, opened 2026-09-08
fix(langchain): do not fail the caller when the callback handler cannot load

jtodaroii:
## Description of changes

Makes the LangChain instrumentation's lazy handler load non-fatal.

`_patchCallbackManager` wraps `CallbackManager._configureSync` and lazily does `require('./callback-handler')` on first use. That handler imports `@langchain/core` directly, so it only resolves when the distro can reach the application's `node_modules`. When the distro is mounted elsewhere, as the OpenTelemetry Operator does at `/otel-auto-instrumentation-nodejs`, the require throws `MODULE_NOT_FOUND` out of `_configureSync` and into the application's request path, failing every model invocation.

This change catches that failure, warns once, latches a `_handlerLoadFailed` flag so the load is not retried on every call, and falls through to the original `_configureSync`. A handler that cannot be loaded now costs tracing instead of the request.

Scope is deliberately narrow: **this is the resilience half only.** It does not change how the handler resolves `@langchain/core`, and it does not address the related problem in `detectConflictingInstrumentation`, which uses `require.resolve` against the same unreachable paths and therefore never detects a conflict in this deployment shape. Both are described in #548. Happy to follow up with the resolution fix separately, since that one is a larger refactor of `callback-handler.ts` and CONTRIBUTING asks that refactors not ride along with functional changes.

Regardless of how resolution is eventually fixed, an observability agent should not be able to fail the request it is observing, so this seemed worth landing on its own.

## Testing

Added `callback handler load failure` to `test/instrumentation/instrumentation-langchain/instrumentation.test.ts`. It forces the handler construction to throw and asserts that the patched `_configureSync` returns the original result, leaves `_handler` undefined, sets `_handlerLoadFailed`, and does not retry on a second call.

Verified the test fails without the source change, reproducing the reported error:

```
1) callback handler load failure
     falls through to the original call and does not retry the failed load:
   Error: Cannot find module '@langchain/core/callbacks/base'
    at Object._configureSync (src\instrumentation\instrumentation-langchain\instrumentation.ts:118:40)
```

With the change, the LangChain suite is green (105 passing). `tsc -p .` is clean and Prettier reports no issues on both touched files.

One note on the wider suite: `test/register.test.ts` shows 26 failures in my local Windows environment, but it shows the identical 26 on an unmodified checkout of `main`, so it is pre-existing here and unrelated to this change.


jtodaroii:
This is waiting on workflow approval rather than on review feedback. Both required runs were created against `8fcadddd` on 2026-09-08 and are still at `action_required`, so CI has never executed:

- `JavaScript Instrumentation PR Build` — https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/34243558214
- `CodeQL Advanced` — https://github.com/aws-observability/aws-otel-js-instrumentation/actions/runs/34243557840

Could a maintainer approve them? First-time-contributor gating means I cannot start them from my side, and without a CI signal there is nothing for a reviewer to act on.

liustve:
Thanks @jtodaroii, closing in favor of: https://github.com/aws-observability/aws-otel-js-instrumentation/pull/553

PR #551, opened 2026-09-10
fix(langchain): remove runtime dependency on @langchain/core

liustve:
## Description of changes

The LangChain callback handler imported `@langchain/core` as **runtime values**:

```ts
import { BaseCallbackHandler } from '@langchain/core/callbacks/base';
import { coerceMessageLikeToMessage, isAIMessage } from '@langchain/core/messages';
```

Bare specifiers resolve by walking *up* from the requiring file. When the distro is
mounted outside the application's `node_modules` — as the OpenTelemetry Operator does,
at `/otel-auto-instrumentation-nodejs` — that walk only reaches
`/otel-auto-instrumentation-nodejs/node_modules`, which doesn't contain `@langchain/core`
(it's a devDependency, never shipped). `/app/node_modules` is a sibling branch and is
never consulted, so the first LangChain call throws `MODULE_NOT_FOUND`.

The module hook is unaffected by this: `require-in-the-middle` intercepts the exports
*the application itself* resolved, which is why the patch applies cleanly even in that
layout (`baseDir: /app/node_modules/@langchain/core` in debug logs). The values we need
are therefore already in hand — they just need to be passed in rather than looked up again.

Both imports are removed:

- `extends BaseCallbackHandler` → `implements CallbackHandlerMethods` (type-only).
  LangChain duck-types handlers via [`isBaseCallbackHandler`](https://github.com/langchain-ai/langchainjs/blob/0c799481f691e046a4533588fc96e190669fa16e/libs/langchain-core/src/callbacks/base.ts),
  which needs only `copy`, `name` and `awaitHandlers`.
- `coerceMessageLikeToMessage` / `isAIMessage` now come from the application's own
  `@langchain/core/dist/messages/index` exports, via an added `InstrumentationNodeModuleFile`.

Every remaining `@langchain/core` import is `import type` and erases at compile time, so
the instrumentation no longer resolves the package at runtime at all. This mirrors upstream
`instrumentation-bunyan` (duck-typed handler class rather than extending the instrumented
package's base class) and `instrumentation-pino` (`moduleExports` passed in rather than
required).

Because there is no longer anything that can throw, this needs no `try`/`catch` or
disable-on-failure latch, and the instrumentation stays **fully functional** under the
operator rather than silently switching off. Alternative to #549.

Only `instrumentation-langchain` was affected: `instrumentation-openai-agents` imports
`@openai/agents` as `import type` only, and `instrumentation-vercel-ai` imports nothing
from `ai` at all.

## Testing

Unit: `1782 passing`, 0 failing (104 in `instrumentation-langchain`). `tsc` clean, Prettier
clean, eslint 0 errors.

End-to-end on EKS with the `amazon-cloudwatch-observability` add-on, with the distro
dependency **removed** from the application images so the operator's
`/otel-auto-instrumentation-nodejs` mount is the only copy — i.e. the exact failing layout.
A/B against the official `adot-autoinstrumentation-node:v0.12.0`:

| `POST /invocations` | v0.12.0 | this change |
|---|---|---|
| langchain | **500** `Cannot find module '@langchain/core/callbacks/base'` | **200** |
| openai-agents | 200 | 200 |
| vercel-ai | 200 | 200 |

Full GenAI trees reach X-Ray — `invoke_agent LangGraph` → 6 × `execute_tool` →
Bedrock/DynamoDB/S3 — with `gen_ai.usage.input_tokens`, `gen_ai.response.id` and
`gen_ai.tool.call.id` populated, confirming the helpers were genuinely supplied through the
hook rather than left `undefined`.

## Follow-up (not in this PR)

`detectConflictingInstrumentation` (`src/utils.ts`) has the same root cause — it calls
`require.resolve(pkg)`, which resolves from the distro's directory, so under the operator it
can never see a competing instrumentation installed in the application's tree and always
fails open. Tracked separately in #548.


aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-langchain/instrumentation.ts
```diff
@@ -96,11 +96,30 @@ export class LangChainInstrumentation extends InstrumentationBase<LangChainInstr
...
+          // Not a patch: hands the callback handler the application's own message helpers, so that
+          // it never has to require('@langchain/core') from a path the distro cannot resolve.
+          ...moduleFiles(
+            '@langchain/core/dist/messages/index',
+            (m: any) => {
+              this._setLangChainMessagesModule(m);
+              return m;
+            },
+            (m: any) => {
+              this._setLangChainMessagesModule(undefined);
```
jj22ee: nit: since nothing is wrapped here, does the unpatch need to clear these at all? They are just the app's own exports so they stay valid either way, and clearing them can strip the helpers from a still-enabled instance.

aws-distro-opentelemetry-node-autoinstrumentation/src/instrumentation/instrumentation-langchain/callback-handler.ts
```diff
@@ -37,40 +37,59 @@ import { contentToParts, PROVIDER_MAP, serializeToJson, toToolAttributeValue } f
...
+// Supplied by the instrumentation from the application's own module exports, via the hook. Nothing
+// here may import @langchain/core at runtime: the distro cannot resolve the application's
+// node_modules when mounted elsewhere, as the OpenTelemetry Operator does.
+let coerceMessageLikeToMessage: ((message: BaseMessageLike) => BaseMessage) | undefined;
+let isAIMessage: ((message: BaseMessage) => message is AIMessage) | undefined;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function setLangChainMessagesModule(messagesModule: any): void {
+  coerceMessageLikeToMessage = messagesModule?.coerceMessageLikeToMessage;
+  isAIMessage = messagesModule?.isAIMessage;
```
jj22ee: nit: can we add a warn/debug if these are still undefined at use time? Would help if a 1.x minor moves `dist/messages/index`, since the manager patch would still apply and the attributes would just silently drop.

jj22ee (approved):

wangzlei (approved):

